PR #97, opened 2023-10-16
Delete manufacturer #72

### LINKED ISSUE #72: As an admin, I want to be able to delete manufacturers
- [ ] Delete a Manufacturer ( on manufacturer tab )
    -  [ ] Implement API delete request 
    -  [ ] Implement UI design (actions column on the left hand side like catalogue items table)
    -  [ ] Implement error checking

MatteoGuarnaccia5:
## Description

Enables users to delete manufacturers. Manufacturers that are a part of a Catalogue Item cannot be deleted and if the user tries to an appropriate error message will be shown

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Check that deleting a manufacturer that is part of an catalogue item is handled

## Agile board tracking

closes #72


joelvdavies (changes requested):
A couple of minor things, and some missing coverage. There are also some conflicts, and add is broken but suspect that would be fixed by merging develop in.

src/manufacturer/manufacturer.component.test.tsx
```diff
@@ -82,4 +82,24 @@ describe('Manufacturer', () => {
...
+    );
+
+    expect(screen.getByText('Delete Manufacturer')).toBeInTheDocument();
+
+    const closeButton = screen.getByRole('button', { name: 'Cancel' });
+    await user.click(closeButton);
+
+    await waitFor(() => {
+      expect(screen.getByText('Actions')).toBeInTheDocument();
+    });
```
joelvdavies, line 103: Could this instead be something like
```suggestion
    await waitFor(() => {
      expect(
        screen.queryByText('Delete Manufacturer')
      ).not.toBeInTheDocument();
```
The 'Actions' heading will be in the document regardless of whether it closes or not.

joelvdavies (approved):

PR #112, opened 2023-10-30
Edit manufacturers #73

### LINKED ISSUE #73: As an operator/admin, I want to be able to edit manufacturers
- [ ] Edit a Manufacturer ( on manufacturer tab )
    -  [ ] Implement API patch request 
    -  [ ] Implement UI design (actions column on the left hand side like catalogue items table)
    -  [ ] Implement error checking

MatteoGuarnaccia5:
## Description

Edit a manufacturer logic. Dialog opens when edit icon clicked, which is prepopulated. Users can edit fields and sends patch request. Error handling in place to ensure name cannot be edited to one that already exists in database, that at least one change is made, and that required fields are not left blank

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

closes #73


src/app.types.tsx
```diff
@@ -32,32 +32,39 @@ export interface CatalogueCategory {
...
 }
+export interface AddManufacturer {
+  name: string;
+  url?: string;
+  address: Address | undefined;
+  telephone?: string;
+}
 
-export interface Manufacturer {
+export interface AddManufacturerResponse {
```
joelvdavies, line 42:
```suggestion
export interface Manufacturer {
```
Shouldn't this replace the Manufacturer below? It is what is returned when getting or doing the post.

src/manufacturer/manufacturer.component.tsx
```diff
@@ -151,6 +161,11 @@ function ManufacturerComponent() {
                       <IconButton
                         size="small"
                         aria-label={`Edit ${item.name} manufacturer`}
+                        onClick={() => {
+                          setEditManufacturerDialogOpen(true);
+                          setSelectedManufacturer(item);
+                          setManufacturer(item);
+                        }}
```
joelvdavies, line 168: Same as https://github.com/ral-facilities/inventory-management-system/pull/97#discussion_r1386439036 could you add test that covers the opening an closing of the edit dialog to cover these lines.

src/api/manufacturer.tsx
```diff
@@ -104,3 +106,72 @@ export const useDeleteManufacturer = (): UseMutationResult<
     },
   });
 };
+
+const fetchManufacturerById = async (
```
joelvdavies, line 110:
```suggestion
const fetchManufacturer = async (
```

src/api/manufacturer.tsx
```diff
@@ -104,3 +106,72 @@ export const useDeleteManufacturer = (): UseMutationResult<
...
+  if (settingsResult) {
+    apiUrl = settingsResult['apiUrl'];
+  }
+
+  return axios.get(`${apiUrl}/v1/manufacturers/${id}`, {}).then((response) => {
+    return response.data;
+  });
+};
+
+export const useManufacturerById = (
```
joelvdavies, line 125:
```suggestion
export const useManufacturer = (
```

src/api/manufacturer.tsx
```diff
@@ -104,3 +106,72 @@ export const useDeleteManufacturer = (): UseMutationResult<
...
+  id: string | undefined
+): UseQueryResult<ManufacturerDetail, AxiosError> => {
+  return useQuery<ManufacturerDetail, AxiosError>(
+    ['ManufacturerByID', id],
```
joelvdavies, line 129:
```suggestion
    ['Manufacturer', id],
```

src/mocks/handlers.ts
```diff
@@ -170,6 +180,38 @@ export const handlers = [
...
+      return res(
+        ctx.status(409),
+        ctx.json({
+          detail:
+            'A manufacturer with the same name has been found. Please enter a different name',
+        })
+      );
+    }
+    if (body.name === 'Error 500') {
+      return res(ctx.status(500), ctx.json(''));
```
joelvdavies, line 196: This line is never run in unit tests, I think we need a test for the catchAllError requests the user to refresh like the other dialogs.

Similarly https://github.com/ral-facilities/inventory-management-system/pull/112/files#diff-bb09d52c957fd04a8ed1a9f64c0cd6da4bb6c207f92a0fa1bb81a0e957aeaf49L129 is also not run, but I think it can just be removed?

joelvdavies (changes requested):
Worked well for me, just some comments on the code and unit test coverage. A few things I will leave for edit to avoid more conflicts.

src/app.types.tsx
MatteoGuarnaccia5: the interface does not have code though, a post request returns a manufacturer that includes a code. Maybe that was my mistake but that's why they are separate interfaces

src/mocks/handlers.ts
MatteoGuarnaccia5: Yes it can be removed. And I'll add a test to check for catchAllError requests

src/app.types.tsx
joelvdavies: The responses of the post, patch or get request is all the same object, they each have code regardless just the one below doesn't (which wont error even though it is being returned by the api), so it can be replaced with one that does include the code.

src/manufacturer/manufacturerDialog.component.tsx
```diff
@@ -160,19 +205,158 @@ function AddManufacturerDialog(props: AddManufacturerDialogProps) {
...
+
+      if (isNameUpdated) {
+        ManufacturerToEdit = {
+          ...ManufacturerToEdit,
+          name: manufacturer.name,
+        };
+      }
+      if (isURLUpdated) {
+        ManufacturerToEdit = {
+          ...ManufacturerToEdit,
```
joelvdavies, line 254: Ah missed that this wasn't covered either, it seems there should be a unit test that checks the url can be updated.

src/manufacturer/manufacturerDialog.component.tsx
```diff
@@ -160,19 +205,158 @@ function AddManufacturerDialog(props: AddManufacturerDialogProps) {
...
+      let ManufacturerToEdit: EditManufacturer = {
+        id: selectedManufacturer?.id,
+      };
+
+      if (isNameUpdated) {
+        ManufacturerToEdit = {
+          ...ManufacturerToEdit,
+          name: manufacturer.name,
+        };
+      }
```
joelvdavies, line 251: Just spoken to Joshua, these could also be simplified slightly
```suggestion
     isNameUpdated && (ManufacturerToEdit .name = manufacturer.name)
```
But it wont work for the address as its nested. Best to address this on  #118 though.

src/manufacturer/manufacturerDialog.component.tsx
joelvdavies: There is a test, just I think the url given to it is "Test", so its invalid which means this code doesn't run, so I think it just needs to be changed to a valid one. (https://github.com/ral-facilities/inventory-management-system/pull/112/files#diff-afc4e4a1b721a32857daaa89d191c8b99a161a5acefc3737abbdc4b8cdf69e41R583)

src/manufacturer/manufacturerDialog.component.tsx
MatteoGuarnaccia5: Addressed in #118 with commit "simplified edit #73"

src/manufacturer/manufacturerDialog.component.test.tsx
```diff
@@ -0,0 +1,755 @@
...
+      expect(
+        screen.getByText('Please enter a street name.')
+      ).toBeInTheDocument();
+      expect(
+        screen.getByText('Please enter a post code or zip code.')
+      ).toBeInTheDocument();
+      expect(onClose).not.toHaveBeenCalled();
+    });
+
+    it.only('CatchAllError request works correctly and displays refresh page message', async () => {
```
joelvdavies, line 555:
```suggestion
    it('CatchAllError request works correctly and displays refresh page message', async () => {
```

src/api/manufacturer.tsx
```diff
@@ -11,6 +11,8 @@ import { settings } from '../settings';
 import {
   AddManufacturer,
   AddManufacturerResponse,
+  EditManufacturer,
+  // ManufacturerDetail,
```
joelvdavies, line 15:
```suggestion
```

joelvdavies (approved):

joelvdavies (approved):

PR #118, opened 2023-11-02
Change address fields #117

### LINKED ISSUE #117: Address input field to become international standard
Change the address field for manufacturers so that international address can also be inputted by users, and are accurate in representing the address of a manufacturer to other users. 

image

Above is a good example. However ours will have only one address line and use 'town', and 'county'

MatteoGuarnaccia5:
## Description

Changed address fields so that it can be used internationally. Updated tests accordingly

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
to be merged in only when https://github.com/ral-facilities/inventory-management-system-api/pull/85#top merged in 
and to be merged in at the same time as https://github.com/ral-facilities/inventory-management-system-api/pull/108
closes #117


src/app.types.tsx
```diff
@@ -121,19 +122,19 @@ export interface ErrorParsing {
...
-  building_number?: string;
-  street_name?: string;
-  town?: string;
-  county?: string;
-  postcode?: string;
+  address_line: string | null;
+  town: string | null;
+  county: string | null;
+  postcode: string | null;
+  country: string | null;
```
joelvdavies, line 137: These should also have an `?` and only have `| null` if they are actually optional.

joelvdavies (changes requested):
A few things I noticed right now, with the types, currently if something is empty in the add dialogue e.g. the town it sends null
image

Ideally this would instead use undefined in typescript so that it isn't sent at all in the request. Its the same sort of thing with the edit requests.

The address display also needs some sort of logic when  rendering as null fields appear like
image

src/app.types.tsx
```diff
@@ -32,36 +32,37 @@ export interface CatalogueCategory {
   is_leaf: boolean;
   catalogue_item_properties?: CatalogueCategoryFormData[];
 }
+
 export interface AddManufacturer {
```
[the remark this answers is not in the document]
joelvdavies: You can also use what Joshua has here https://github.com/ral-facilities/inventory-management-system/blob/afcb642a56f54c0bd0a7238ec86e24355337ff92/src/app.types.tsx#L109 for edit, using `Partial` to effectively copy everything in the `CatalogueItemDetails` but allowing them to be undefined as well. This would also mean adding a separate Details that is extended by the others.

src/manufacturer/manufacturerDialog.component.tsx
```diff
@@ -222,53 +211,34 @@ function ManufacturerDialog(props: ManufacturerDialogProps) {
...
       const isCountyUpdated =
         manufacturer.address?.county !== selectedManufacturer.address.county;
-      const ispostcodeUpdated =
+      const isPostcodeUpdated =
         manufacturer.address?.postcode !==
         selectedManufacturer.address.postcode;
+      const isCountryUpdated =
+        manufacturer.address?.country !== selectedManufacturer.address.country;
       const isTelephoneUpdated =
         manufacturer.telephone !== selectedManufacturer.telephone;
```
joelvdavies, line 227: I think these may need modifying - where they are optional, the types will always be different so the result will always be true e.g. `selectedManufacturer.telephone` can be None, but `manufacturer.telephone` would be "" if I added then deleted something again. It could be a case of adding `value || null` on the onChange methods so they are stored as the same types, which should hopefully also resolve them being added to the address data sent as currently changing the country on its own sends
```
{"address":{"address_line":"Address Line","town":null,"county":null,"country":"Country3","postcode":"Postcode"}}
```
But ideally it would just be
```
{"address":{"country":"Country3"}}
```

joelvdavies (changes requested):
A couple more type related things. Also could the order of the fields be modified in the dialog/when displaying to have country after county, instead of before the first address line
image
as that is more standard.

src/app.types.tsx
```diff
@@ -32,36 +32,37 @@ export interface CatalogueCategory {
...
   name: string;
   url?: string;
-  address: Address | undefined;
+  address?: Address;
   telephone?: string;
 }
 
 export interface AddManufacturerResponse {
```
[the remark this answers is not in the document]
MatteoGuarnaccia5: the Manufacturer type has an ID though?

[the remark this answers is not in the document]
joelvdavies: Its the same as https://github.com/ral-facilities/inventory-management-system/pull/112#discussion_r1387925427, it is already returning an id but it wont error if its attempting to cast to this type, it will just ignore it so changing to Manufacturer will just pick it up again.

src/app.types.tsx
```diff
@@ -32,36 +32,30 @@ export interface CatalogueCategory {
   is_leaf: boolean;
   catalogue_item_properties?: CatalogueCategoryFormData[];
 }
-export interface AddManufacturer {
-  name: string;
-  url?: string;
-  address: Address | undefined;
-  telephone?: string;
-}
```
joelvdavies, line 40: Ah sorry I meant remove AddManufacturerResponse, not AddManufacturer, github showed both the line I selected and several lines above it so it looked like I meant remove this.

src/manufacturer/manufacturerDialog.component.tsx
```diff
@@ -396,54 +369,52 @@ function ManufacturerDialog(props: ManufacturerDialogProps) {
...
+          </Grid>
+          <Grid item>
             <TextField
-              label="Building number"
-              required={type === 'create' ? true : false}
+              label="Country"
+              required={true}
               sx={{ marginLeft: '4px', my: '8px' }} // Adjusted the width and margin
-              value={manufacturer.address.building_number}
+              value={manufacturer.address.country}
```
joelvdavies, line 378: Can this country field be moved below county as well to match the order.

joelvdavies (changes requested):
Sorry I don't think the comment I made was clear on what to remove.

joelvdavies (dismissed):

joshuadkitenge (changes requested):
You currently have two useState for setting the manufacturer details in the manufacturer component . You should only need one useState. (setManufacturer and setSelectedManufacturer)

src/app.types.tsx
```diff
@@ -38,36 +38,31 @@ export interface CatalogueCategory {
...
-  id?: string;
+  telephone?: string | null;
+  id?: string | null;
 }
 
-export interface Manufacturer {
-  id?: string;
+export interface ManufacturerDetails {
   name: string;
   url?: string;
```
joelvdavies, line 59:
```suggestion
  url?: string | null;
```
In fact testing it, it was null in my case so the type needs to be changed to this regardless.

src/app.types.tsx
```diff
@@ -38,36 +38,31 @@ export interface CatalogueCategory {
...
-  address: Address;
-  telephone: string;
-  code: string;
+  address: AddAddress;
+  telephone?: string | null;
 }
 
 export interface EditManufacturer {
   name?: string;
   url?: string;
```
joelvdavies, line 51: Same here, could be
```suggestion
  url?: string | null;
```

src/app.types.tsx
```diff
@@ -38,36 +38,31 @@ export interface CatalogueCategory {
   is_leaf: boolean;
   catalogue_item_properties?: CatalogueCategoryFormData[];
 }
+
 export interface AddManufacturer {
   name: string;
   url?: string;
```
joelvdavies, line 44: I double checked, it was fixed here https://github.com/ral-facilities/inventory-management-system-api/commit/f576bbf01d8b4cd708696e85e019c26fb7db0f4a so could technically be `| null` if we wanted now.
```suggestion
  url?: string | null;
```

joelvdavies (commented):
One last note on the types - also speaking to Joshua, we need to merge
https://github.com/ral-facilities/inventory-management-system-api/pull/115 and https://github.com/ral-facilities/inventory-management-system/pull/128 first as the user feedback is coming up.

Also saw this which is a good sign
image

joelvdavies (approved):

joshuadkitenge (approved):

PR #128, opened 2023-11-09
Add addtional fields to catalogue items #107

### LINKED ISSUE #107: Add addtional fields for catalogue item

joshuadkitenge:
## Description

-  Add all the additional field to catalogue items.
     - Edit Catalogue item
     - Landing page
     - Add catalogue item
     - Table View 
     

## Testing instructions

- [ ] Check the CRU functionality still works as expected
- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #107


src/app.types.tsx
```diff
@@ -47,8 +47,34 @@ export interface CatalogueCategoryFormData {
...
-  description: string | undefined;
+  name: string;
+  description: string | null;
+  cost_gbp: number;
+  cost_to_rework_gbp: number | null;
+  days_to_replace: number;
+  days_to_rework: number | null;
+  drawing_number: string | null;
+  drawing_link: string | null;
+  model_number: string | null;
```
joelvdavies, line 58: This was changed to `item_model_number` due to a pydantic conflict.

src/app.types.tsx
```diff
@@ -68,25 +94,19 @@ export interface CatalogueItemPropertyResponse {
   unit: string;
 }
 
-export interface CatalogueItem {
+export interface CatalogueItem extends CatalogueItemDetails {
   catalogue_category_id: string;
-  name: string | undefined;
-  description: string;
   properties: CatalogueItemPropertyResponse[];
   manufacturer: CatalogueItemManufacturer;
```
joelvdavies, line 100: I think this could go in CatalogueItemDetails as its common, but then again it would really be the ID, which isn't in yet so I would move it in that instead.

src/app.types.tsx
```diff
@@ -47,8 +47,34 @@ export interface CatalogueCategoryFormData {
...
+export interface CatalogueDetailsErrorMessages {
+  name: string;
+  description: string;
+  cost_gbp: string;
+  cost_to_rework_gbp: string;
+  days_to_replace: string;
+  days_to_rework: string;
+  drawing_number: string;
+  drawing_link: string;
+  model_number: string;
```
joelvdavies, line 77:
```suggestion
  item_model_number: string;
```

joelvdavies (commented):
As mentioned just a note on model_number, and probably needs develop merged for `web_url` -> `url`. I also need to update my generation script so I will do that before I review fully.

src/catalogue/items/catalogueItemsLandingPage.component.tsx
```diff
@@ -18,6 +18,7 @@ import { useCatalogueCategoryById } from '../../api/catalogueCategory';
 import CatalogueItemsDialog from './catalogueItemsDialog.component';
 import {
   CatalogueItemDetails,
```
joelvdavies, line 20:
```suggestion
```
I don't think this is used.

src/app.types.tsx
```diff
@@ -68,25 +94,19 @@ export interface CatalogueItemPropertyResponse {
...
+export interface CatalogueItem extends CatalogueItemDetails {
   catalogue_category_id: string;
-  name: string | undefined;
-  description: string;
   properties: CatalogueItemPropertyResponse[];
   manufacturer: CatalogueItemManufacturer;
   id: string;
 }
-export interface AddCatalogueItem {
+export interface AddCatalogueItem extends CatalogueItemDetails {
```
joelvdavies, line 103: Happy to leave it for now, but I think it makes more sense the optional fields here to be undefined instead of null as the backend will take care of the defaults, then EditCatalogueItem can extend `Partial<AddCatalogueItem>` to make the remaining fields optional. This would mean CatalogueItem couldn't inherit from the same thing, and CatalogueItemDetailsPlaceholder would need to be | undefined as well though.

src/catalogue/items/catalogueItemsTable.component.tsx
```diff
@@ -303,6 +429,58 @@ const CatalogueItemsTable = (props: CatalogueItemsTableProps) => {
...
+                    {item.obsolete_replacement_catalogue_item_id ? (
+                      <MuiLink
+                        underline="hover"
+                        component={Link}
+                        to={`items/${item.obsolete_replacement_catalogue_item_id}`}
+                      >
+                        Click here
+                      </MuiLink>
+                    ) : (
+                      'None'
```
joelvdavies, line 459: Not sure this should display as None as the others don't - I don't know what it would look like in the datagrid though
image

src/catalogue/items/catalogueItemsDialog.component.tsx
```diff
@@ -468,14 +534,61 @@ function CatalogueItemsDialog(props: CatalogueItemsDialogProps) {
...
+          drawing_link: details.drawing_link,
+        };
+      }
+
+      if (isModelNumberUpdated) {
+        catalogueItem = {
+          ...catalogueItem,
+          model_number: details.model_number,
         };
       }
```
joelvdavies, line 593: These can be simplified like the name was.

src/catalogue/items/catalogueItemsDialog.component.test.tsx
```diff
@@ -865,14 +975,28 @@ describe('Catalogue Items Dialog', () => {
       createView();
 
       const nameInput = screen.getByLabelText('Name *');
-      await user.type(nameInput, newName);
+      fireEvent.change(nameInput, { target: { value: newName } });
```
joelvdavies, line 978: Why did this change?

src/catalogue/items/catalogueItemsLandingPage.component.tsx
```diff
@@ -112,55 +148,203 @@ function CatalogueItemsLandingPage() {
...
-                  <ExpandMoreIcon />
-                </>
-              )}
-            </Box>
-            <Collapse in={showProperties}>
-              <Grid container spacing={1}>
-                {catalogueItemIdData.properties &&
-                  catalogueItemIdData.properties.map((property, index) => (
-                    <Grid item xs={12} sm={6} md={4} key={index}>
+          <Grid container spacing={1} flexDirection="column">
```
joelvdavies, line 151: Might be related to my generated data, but my landing page here has a strange horizonatal size causing a scrollbar when I think it should be able to fit
image

src/catalogue/items/catalogueItemsDialog.component.tsx
```diff
@@ -300,9 +414,30 @@ function CatalogueItemsDialog(props: CatalogueItemsDialogProps) {
     ) as CatalogueItemProperty[];
 
     const catalogueItem: AddCatalogueItem = {
+      name: catalogueItemDetails.name ?? '',
+      cost_gbp: catalogueItemDetails.cost_gbp
+        ? Number(catalogueItemDetails.cost_gbp)
+        : 0,
```
joelvdavies, line 420: Could this be covered in the unit tests - there are only two that aren't.

src/catalogue/items/catalogueItemsDialog.component.tsx
```diff
@@ -300,9 +414,30 @@ function CatalogueItemsDialog(props: CatalogueItemsDialogProps) {
...
+      cost_to_rework_gbp:
+        catalogueItemDetails.cost_to_rework_gbp === null
+          ? null
+          : Number(catalogueItemDetails.cost_to_rework_gbp),
+      days_to_replace: catalogueItemDetails.days_to_replace
+        ? Number(catalogueItemDetails.days_to_replace)
+        : 0,
```
joelvdavies, line 427: Could this be covered in the unit tests - this is the only other missing line.

src/catalogue/items/catalogueItemsDialog.component.tsx
```diff
@@ -315,143 +450,76 @@ function CatalogueItemsDialog(props: CatalogueItemsDialogProps) {
...
-        setManufacturerAddressError(true);
-        hasErrors = true;
+      if (hasErrors) {
+        return; // Do not proceed with saving if there are errors
       }
+      const details = {
+        name: catalogueItemDetails.name ?? '',
+        cost_gbp: catalogueItemDetails.cost_gbp
+          ? Number(catalogueItemDetails.cost_gbp)
+          : 0,
```
joelvdavies, line 471: I think this would be covered if the other lines were.

src/catalogue/items/catalogueItemsDialog.component.tsx
```diff
@@ -315,143 +450,76 @@ function CatalogueItemsDialog(props: CatalogueItemsDialogProps) {
...
+        cost_to_rework_gbp:
+          catalogueItemDetails.cost_to_rework_gbp === null
+            ? null
+            : Number(catalogueItemDetails.cost_to_rework_gbp),
+        days_to_replace: catalogueItemDetails.days_to_replace
+          ? Number(catalogueItemDetails.days_to_replace)
+          : 0,
```
joelvdavies, line 478: I think this would be covered if the other lines were.

src/catalogue/items/catalogueItemsDialog.component.tsx
```diff
@@ -315,143 +450,76 @@ function CatalogueItemsDialog(props: CatalogueItemsDialogProps) {
...
+            ? null
+            : Number(catalogueItemDetails.days_to_rework),
+        description: catalogueItemDetails.description,
+        item_model_number: catalogueItemDetails.item_model_number,
+        is_obsolete: false,
+        obsolete_reason: catalogueItemDetails.obsolete_reason,
+        obsolete_replacement_catalogue_item_id:
+          catalogueItemDetails.obsolete_replacement_catalogue_item_id,
+        drawing_link: catalogueItemDetails.drawing_link,
+        drawing_number: catalogueItemDetails.drawing_number,
```
joelvdavies, line 490: Could these be the same order as in the app.types.

joelvdavies (commented):
A few more minor things, the code LGTM, ~will test it properly tomorrow~.

src/catalogue/items/catalogueItemsDialog.component.tsx
joshuadkitenge: These are impossible to trigger because of the error checking

src/catalogue/items/catalogueItemsLandingPage.component.tsx
joshuadkitenge: this is because of the manufacturer address. That can't be fix until matteo pr is in by adding line breaks  or by reordering the manufacturer details

src/catalogue/items/catalogueItemsDialog.component.test.tsx
joshuadkitenge: Sometimes userEvent doesn't work as expected.  Sometimes the onChangeValue changes one character at a time (instead just of "test" it would be called with "t","e","s" and "t") so you can't do the assertion check. For now I just change it to fireEvent because that works but I plan on investigated this in the future.

src/catalogue/items/catalogueItemsLandingPage.component.tsx
joshuadkitenge: also description should be centered. I'll investigate that as well

joshuadkitenge: The issue was because I used margin instead of padding

joelvdavies (approved):

VKTB (approved):

PR #134, opened 2023-11-15
Edit systems #17

### LINKED ISSUE #17: As an operator/admin, I want to be able to edit systems
- [ ] Implement User Interface and API request for editing systems (2 weeks)
- [ ] Implement Button to edit system
- [ ] Implement window with pre-populated fields 
- [ ] Implement post request to add to DB

### LINKED ISSUE #146: System importance chip colour wrong in SciGateway
Believe its the same as https://github.com/mui/material-ui/issues/31857. Have currently used the `color` prop directly, but really need to use `sx` to override. At the moment the colour in scigateway is overridden specifically for their usage in datagateway to meet contrast requirements for accessibility, it may not be needed anymore with MUIv5.

joelvdavies:
## Description

Adds the ability to edit a system

image

Currently I just put the button next to delete, hoping it will be clearer where these should go once I have looked into moving systems, particularly since that will probably involve selecting more than one system at once.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

Closes #17, Closes #146


VKTB (commented):
I noticed a couple of things while testing this.
- If I only edit the name of the system then the changes are updated immediately on the page but this is not the case when I try to edit any of the other fields.
- If I remove all the field inputs from the Edit System window, click Save, click Cancel, and then click Edit again, the field inputs are empty. Shouldn't the inputs be shown given I have exited the edit window and opened again?

joelvdavies:
> I noticed a couple of things while testing this.
> 
> * If I only edit the name of the system then the changes are updated immediately on the page but this is not the case when I try to edit any of the other fields.
😄 that was tricky to debug - the breadcrumbs were causing a re-render of the page which meant the removeQueries functioned as I thought it would, but it needed to be invalidateQueries to actually work properly.

> * If I remove all the field inputs from the Edit System window, click Save, click Cancel, and then click Edit again, the field inputs are empty. Shouldn't the inputs be shown given I have exited the edit window and opened again?
Yeah I agree - I was in two minds at the time - what if someone partially edited then closed the window?, but cancel implies undoing the changes so I have made it reset.

VKTB:
@joelvdavies Thanks for taking a look and fixing these. They are working as expected now. However, I think the changes have affected the adding of systems in some way because when I created a test system, I had to refresh the page to see it in the list. I don't remember having to do this earlier today when I again had to create a test system, do you mind having a look?

joelvdavies:
> @joelvdavies Thanks for taking a look and fixing these. They are working as expected now. However, I think the changes have affected the adding of systems in some way because when I create a test system, I have to refresh the page to see it in the list. I don't remember having to do this earlier today when I again had to create a test system. Do you mind having a look, please?

🤦‍♂️ - Thanks for spotting that, I made a change to the invalidation to only invalidate the systems within the parent, but I had the wrong type as we use a string "null" meaning it never invalidated if adding in the root.

VKTB (commented):
I noticed something else while testing earlier. If I edit the name of a root system, its name and the breadcrumb get updated but if I navigate back to the root of the systems then the name is still the old one. I also noticed that if I go in its children systems, the relevant breadcrumb part has the updated name of the parent system in some of the children systems but not in all.

joelvdavies:
> I noticed something else while testing earlier. If I edit the name of a root system, its name and the breadcrumb get updated but if I navigate back to the root of the systems then the name is still the old one. I also noticed that if I go in its children systems, the relevant breadcrumb part has the updated name of the parent system in some of the children systems but not in all.

@VKTB I *think* I have addressed those now - tricky to address the latter without invalidating all the breadcrumbs to ensure all children are also invalidated so that is what I have done.

Issue #130, opened 2023-11-13
A user should be a to set an catalogue item as Obsolete


joshuadkitenge:
- Create an obsolete dialogue 
- use MUI stepper to allow user to set is_obsolete, obsolete replacement catalogue item and  obsolete reason
- create pages for  is_obsolete, obsolete replacement catalogue item and and obsolete reason

PR #142, opened 2023-11-18
Obsolete dialog #130

joshuadkitenge:
## Description

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

connect to {issue number}


joshuadkitenge:
LGTM

VKTB:
I think I would prefer the list of catalogue items from the catalogue category I am in to be shown in the 3rd step (_Obsolete Replacement_) rather than the root of catalogue categories. I would expect the replacement catalogue item to likely be in the same catalogue category but if not then I can use the breadcrumbs. Just saves some navigating through if there are lots of subcategories. This is just a comment but looks good to me otherwise.

VKTB (approved):

VKTB (changes requested):
I tested this and noticed the following:
- The _Obsolete replacement link_ column is not updated when I select an obsolete replacement 
- If in the 3rd step (_Obsolete Replacement_), I navigate to root, click _Cancel_. and then come back to the same step for the same catalogue item, the table does not reset as it is still at root.

VKTB (approved):

PR #176, opened 2023-12-11
Copy systems #136

### LINKED ISSUE #136: A user/operator should be able to copy a system
Similar to #100, but for systems - want to be able to copy one or more systems.

joelvdavies:
## Description

Adds the ability to copy one or more systems to another system.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

Closes #136

### DIFF
--- a/cypress/e2e/system.cy.ts
+++ b/cypress/e2e/system.cy.ts
@@ -376,7 +376,7 @@ describe('System', () => {
   });
 
   it('copies systems', () => {
-    cy.visit('/inventory-management-system/systems');
+    cy.visit('/systems');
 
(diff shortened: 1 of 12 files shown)


joshuadkitenge (changes requested):
Functionality works well. Just a few comments. 

- can you add a Copy warning message. That warns users that the subs systems are not copied as well 

- you should be able to do nested copies 



systems copy valid copy

joshuadkitenge (changes requested):
I noticed that the catalogue and systems moved to and copy to success messages are different can you change so that they are the same please

joshuadkitenge (approved):

PR #144, opened 2023-11-23
Integrate manufacturer into catalogue items v2 #108

MatteoGuarnaccia5:
## Description

Integrated manufacturer into catalogue items, changed dialog so that users are able to select a manufacturer from a autocomplete dropdown. they can also add a new manufacturer, which will then be added to dropdown list. The ID of the manufacturer is then passed through to the backend to be added to the catalogue item. Currently the table view displays all the fields of a manufacturer (name, address ,url, telephone number) but this is subject to change based upon CLFs feedback

Note - #132 merged into this PR. So this PR will essentially close it and the issue for it

## Testing instructions

Add a set up instructions describing how the reviewer should test the code

- [ ] Review code
- [ ] Check Actions build
- [ ] Review changes to test coverage

## Agile board tracking

closes #108
closes #129

PR #177, opened 2023-12-11
Move to catalogue item #173

### LINKED ISSUE #173: Move to Catalogue item

joshuadkitenge:
## Description

A user should be able to move a catalogue item from one category to another when the catalogue properties are exactly the same. They can do this by multi selecting catalogue items and clicking the move to button. Then they can navigating to the directory and move the catalogue items if the  catalogue are allowed to move it closes the dialogue if not it displays an error message  

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Move a catalogue item
- [ ] Move error states
- [ ] {more steps here}

## Agile board tracking

closes #173


src/api/catalogueItem.tsx
```diff
@@ -191,3 +194,74 @@ export const useEditCatalogueItem = (): UseMutationResult<
...
+    const promises = MoveToCatalogueItem.catalogueItems.map(
+      async (item: EditCatalogueItem, index) => {
+        const { name, ...itemWithoutName } = item;
+
+        if (
+          MoveToCatalogueItem.selectedItems[index].catalogue_category_id ===
+          item.catalogue_category_id
+        ) {
+          const errorTransferState: TransferState = {
+            name: item.name ?? '',
```
joelvdavies, line 222: If you use the `selectedItems` in the map instead of `catalogueItems` you can avoid the `??` here as it has to be defined.

src/api/catalogueItem.tsx
```diff
@@ -191,3 +194,74 @@ export const useEditCatalogueItem = (): UseMutationResult<
...
+          };
+          transferStates.push(errorTransferState);
+
+          return;
+        }
+
+        return editCatalogueItem(itemWithoutName)
+          .then((result) => {
+            const successTransferState: TransferState = {
+              name: result.name ?? '',
```
joelvdavies, line 235:
```suggestion
              name: result.name,
```
When successful it should be defined.

src/api/catalogueItem.tsx
```diff
@@ -191,3 +194,74 @@ export const useEditCatalogueItem = (): UseMutationResult<
...
+            hasSuccessfulEdit = true;
+          })
+          .catch((error) => {
+            const response = error.response?.data as ErrorParsing;
+
+            const selectedItem = MoveToCatalogueItem.selectedItems.find(
+              (selectedItem) => selectedItem.id === item.id
+            );
+            const errorTransferState: TransferState = {
+              name: selectedItem?.name ?? '',
```
joelvdavies, line 249: If using my above suggestion using `selectedItems` in the map this wont be needed.
```suggestion
            const errorTransferState: TransferState = {
              name: selectedItem.name,
```

cypress/e2e/catalogue/catalogueItems.cy.ts
```diff
@@ -513,4 +513,67 @@ describe('Catalogue Items', () => {
...
+
+    cy.findByRole('dialog')
+      .should('be.visible')
+      .within(() => {
+        cy.contains(
+          'The destination catalogue item properties must precisely match the current destination. Ensure identical attributes, order, and formatting, with no spacing variations.'
+        );
+      });
+  });
+  it('Move here button is disabled when trying to move to the current location', () => {
```
joelvdavies, line 567: I think there should be either an e2e test or unit test to ensure the Move button is also disabled when attempting to move into a catalogue category that isn't a leaf.

joelvdavies (changes requested):
Core functionality worked well for me. Just two minor issues and some suggestions.

cypress/e2e/catalogue/catalogueItems.cy.ts
joshuadkitenge: That is already tested. When I checked if the button is disabled when the dialog is opened because it opens on a non leaf node

src/catalogue/items/catalogueItemsTable.component.tsx
```diff
@@ -3,6 +3,7 @@ import DeleteIcon from '@mui/icons-material/Delete';
 import EditIcon from '@mui/icons-material/Edit';
```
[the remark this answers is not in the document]
joshuadkitenge: you can trigger this error if you navigate to an empty leaf node and then navigate to the parent and navigate back to the empty leaf node the text would be off center. I'm guessing that the error is causing the dialog to rerender causing it to correct itself

cypress/e2e/catalogue/catalogueItems.cy.ts
joelvdavies: Doesn't that just mean its testing that its disabled if moving to the same place, rather than if the category is a leaf?

joshuadkitenge: No, because when a user opens the dialog is opens to the parent catalogue category and not the catalogue category that it is currently in

src/api/catalogueItem.tsx
```diff
@@ -191,3 +194,56 @@ export const useEditCatalogueItem = (): UseMutationResult<
...
+  AxiosError,
+  MoveToCatalogueItem
+> => {
+  const queryClient = useQueryClient();
+  return useMutation(async (moveToCatalogueItem: MoveToCatalogueItem) => {
+    const transferStates: TransferState[] = [];
+    let successfulIds: string[] = [];
+
+    const promises = moveToCatalogueItem.selectedItems.map(
+      async (item: EditCatalogueItem, index) => {
```
joelvdavies, line 209: I believe this type is now `CatalogueItem`

src/api/catalogueItem.tsx
```diff
@@ -191,3 +194,56 @@ export const useEditCatalogueItem = (): UseMutationResult<
...
+        return editCatalogueItem({
+          id: item.id,
+          catalogue_category_id:
+            moveToCatalogueItem.targetCatalogueCategory?.id,
+        })
+          .then((result) => {
+            transferStates.push({
+              name: result.name ?? '',
```
joelvdavies, line 217:
```suggestion
              name: result.name,
```
Since result is now of type CatalogueItem, which always has a name.

src/api/catalogueItem.tsx
```diff
@@ -191,3 +194,56 @@ export const useEditCatalogueItem = (): UseMutationResult<
...
+          });
+      }
+    );
+
+    await Promise.all(promises);
+
+    if (successfulIds.length > 0) {
+      queryClient.invalidateQueries({ queryKey: ['CatalogueItems'] });
+      queryClient.invalidateQueries({ queryKey: ['CatalogueBreadcrumbs'] });
+      successfulIds.map((id: string) =>
```
joelvdavies, line 242:
```suggestion
      successfulIds.forEach((id: string) =>
```
Very minor - but we don't need it to return a new array, so forEach should be slightly faster.

joelvdavies (changes requested):
Very minor comments, will test once more once develop is merged.

src/api/catalogueItem.tsx
```diff
@@ -191,3 +194,56 @@ export const useEditCatalogueItem = (): UseMutationResult<
...
+              state: 'error',
+            });
+          });
+      }
+    );
+
+    await Promise.all(promises);
+
+    if (successfulIds.length > 0) {
+      queryClient.invalidateQueries({ queryKey: ['CatalogueItems'] });
```
joelvdavies, line 240: Also just thought - Should we only invalidate the items needed here? We could do but not sure its worth the effort - the queries will refetch every so often anyway. I only did it for systems as it was simple to do due to the structure, but for catelogue categories it would require something like
image

joshuadkitenge: If you don't invalidate catalogue items the table is not going to update when the catalogue item is moved. I can change it so it only invalidate the queries for the target and the current location as those are the only only that need changing.

joshuadkitenge: image

src/api/catalogueItem.tsx
```diff
@@ -191,3 +194,74 @@ export const useEditCatalogueItem = (): UseMutationResult<
...
+        queryKey: [
+          'CatalogueItems',
+          moveToCatalogueItem.targetCatalogueCategory?.id,
+        ],
+      });
+      // Also need to invalidate each catalogue categories we are moving from (likely just the one)
+      const uniqueCatalogueCategoryIds = new Set(
+        successfulCatalogueCategoryIds
+      );
+      console.log(uniqueCatalogueCategoryIds);
```
joelvdavies, line 253:
```suggestion
```

src/api/catalogueItem.test.tsx
```diff
@@ -183,4 +185,134 @@ describe('catalogue items api functions', () => {
...
+
+  describe('useMoveToCatalogueItem', () => {
+    let props: MoveToCatalogueItem;
+
+    beforeEach(() => {
+      props = {
+        selectedItems: [
+          {
+            catalogue_category_id: '657305a01e468454e97b6389',
+            manufacturer: {
```
MatteoGuarnaccia5, line 197: The manufacturer should be `manufacturer_id` and an Id now that #144 has been merged

src/api/catalogueItem.test.tsx
```diff
@@ -183,4 +185,134 @@ describe('catalogue items api functions', () => {
...
+                name: 'center wavelength',
+                value: 10,
+                unit: 'nm',
+              },
+            ],
+            id: '657305e51e468454e97b638b',
+          },
+          {
+            catalogue_category_id: '657305a01e468454e97b6389',
+            manufacturer: {
```
MatteoGuarnaccia5, line 225: same here

src/catalogue/items/catalogueItemDirectoryDialog.component.test.tsx
```diff
@@ -0,0 +1,257 @@
...
+              unit: 'Joules',
+            },
+            {
+              name: 'Accuracy',
+              value: '±0.5%',
+              unit: '',
+            },
+          ],
+          id: '89',
+          manufacturer: {
```
MatteoGuarnaccia5, line 68: same here it should now be a `manufacturer_id`

src/catalogue/items/catalogueItemDirectoryDialog.component.test.tsx
```diff
@@ -0,0 +1,257 @@
...
+          description: 'Precision energy meters for accurate measurements. 27',
+          properties: [
+            {
+              name: 'Measurement Range',
+              value: 2000,
+              unit: 'Joules',
+            },
+          ],
+          id: '6',
+          manufacturer: {
```
MatteoGuarnaccia5, line 96: same here

MatteoGuarnaccia5 (changes requested):
Works for me, just one thing I noticed in a couple of the test files

MatteoGuarnaccia5 (approved):
LGTM

joelvdavies (approved):

PR #186, opened 2023-12-14
Duplicate property names fix #181

### LINKED ISSUE #181: Duplicate catalogue item properties get ignored when creating a catalogue item
Add error checking to stop users creating catalogue items with duplicate property names

MatteoGuarnaccia5:
## Description

Implemented logic to prevent duplicate catalogue item property names to be used when creating a catalogue category. Check is not case sensitive so will detect words which are the same (regardless of capitalization). Error messages appear on the properties with duplicate names, and all disappear when an value is changed. Logic works for both adding and editing a catalogue category

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] check works for creating a catalogue category
- [ ] check works for editing a catalogue category

## Agile board tracking

closes #181

### DIFF
--- a/src/catalogue/category/catalogueCategoryDialog.component.tsx
+++ b/src/catalogue/category/catalogueCategoryDialog.component.tsx
@@ -72,6 +72,10 @@ const CatalogueCategoryDialog = React.memo(
       undefined
     );
 
+    const [duplicatePropertyError, setDuplicatePropertyError] = React.useState<
+      string[] | undefined
+    >(undefined);
+
     const [catchAllError, setCatchAllError] = React.useState(false);
 
     const { mutateAsync: addCatalogueCategory } = useAddCatalogueCategory();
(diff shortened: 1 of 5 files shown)


src/catalogue/category/catalogueCategoryDialog.component.tsx
```diff
@@ -159,6 +165,29 @@ const CatalogueCategoryDialog = React.memo(
...
+          if (uniqueNames.has(name)) {
+            duplicateNames.push(name);
+          } else {
+            uniqueNames.add(name);
+          }
+        }
+        if (duplicateNames.length > 0) {
+          setDuplicatePropertyError(duplicateNames);
+          hasErrors = true;
+        }
```
joshuadkitenge, line 187: dup item prop name

The property names needs to be trimmed before being checked. Currently if you have "name" and "name " (with a space at the end) they are two different values. same as the back-end change i suggested

joshuadkitenge (changes requested):
Works well just a few minor bugs:

The duplicate name error message should be cleared if the one of the duplicate property rows has been deleted.

currently it persists 

 
dup item prop name delete

joshuadkitenge (approved):
LGTM

PR #205, opened 2024-01-05
Save as system #137

### LINKED ISSUE #137: A user/operator should be able to save a System as
Same as #105 but for systems

joelvdavies:
## Description

- Adds a menu to each system/subsystem with edit, save as and delete (which has been removed from the landing page)
- Adds save as dialog

image
image

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

Closes #137

### DIFF
--- a/cypress/e2e/system.cy.ts
+++ b/cypress/e2e/system.cy.ts
@@ -350,7 +350,7 @@ describe('System', () => {
       });
     });
 
-    it("save as a system eidting only a system's name (in subsystem)", () => {
+    it("save as a system editing only a system's name (in subsystem)", () => {
       cy.visit('/systems/65328f34a40ff5301575a4e3');
 
       cy.findAllByLabelText('Row Actions').eq(0).click();
--- a/cypress/e2e/system.cy.ts
+++ b/cypress/e2e/system.cy.ts
@@ -72,7 +72,11 @@ describe('System', () => {
       }).should(async (postRequests) => {
         expect(postRequests.length).equal(1);
         expect(JSON.stringify(await postRequests[0].json())).equal(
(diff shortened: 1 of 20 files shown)


cypress/e2e/system.cy.ts
```diff
@@ -272,49 +285,148 @@ describe('System', () => {
...
+            location: 'System location',
+            owner: 'System owner',
+            importance: 'medium',
+            parent_id: null,
+          })
+        );
+      });
+    });
+
+    it("save as a system eidting only a system's name (in subsystem)", () => {
```
MatteoGuarnaccia5, line 353:
```suggestion
    it("save as a system editing only a system's name (in subsystem)", () => {
```

MatteoGuarnaccia5 (commented):
image

When moving a system into another system, I found that the chosen system still remains visible, until you refresh the screen or click into the system it was moved into. Should probably refetch the systems after moved so that it displays the correct systems

joshuadkitenge (changes requested):
Everything work well. Just one minor comment:

- can you make it so that when a user "save as" a system that the name defaults with `_copy_{iteration of the copy}`. Just so it matches the implementation done in catalogue items table view.

joelvdavies:
> Everything work well. Just one minor comment:
> 
> * can you make it so that when a user "save as" a system that the name defaults with `_copy_{iteration of the copy}`. Just so it matches the implementation done in catalogue items table view.

Just realised this is actually easier to resolve with the changes to use the table, I will just create conflicts here so I will make a note in the issue and do it there if that's alright (although I have just refactored slightly here to reduce code repetition)

joshuadkitenge (approved):

PR #189, opened 2024-01-03
Disable buttons on error #187

MatteoGuarnaccia5:
## Description

Implemented logic so that save/continue buttons in the dialogs are disabled if there is an error in the dialog.
I only refactored the e2e tests that test for errors, to check if the save/continue button is disabled

## Testing instructions

Add a set up instructions describing how the reviewer should test the code

- [ ] Review code
- [ ] Check Actions build
- [ ] Review changes to test coverage
## Agile board tracking

closes #187

PR #207, opened 2024-01-05
add item #191

### LINKED ISSUE #191: Implement User Interface and API request for adding items
- [ ] Items button/link on the landing page and the table view 
- [ ] Create back button for the landing page and the table view 
- [ ] Implement Add item button and dialog.
- [ ] the add items dialog should have pre-populated values from the catalogue item

joshuadkitenge:
## Description

A user should be able to add an item. A user should also be able to navigate to the items page on the catalogue category table view and the catalogue item landing page. The items page should allow the user to navigate back to either of the pages.  

- ~The items dialog unit tests and the items e2e test  fail locally due the Timezone issues. Ignore this for now. We will deal with this after MVP.~ - Fixed
-  
## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #191


joelvdavies (changes requested):
Mostly just me being pedantic, would be good if the testing issue could fixed though. Worked well when I tested it locally, also checked the behaviour of changing my local time zone which atomically converted the times on the front end, which I think is reasonable. The backend just remains in UTC.

src/items/itemDialog.component.tsx
```diff
@@ -0,0 +1,628 @@
...
+      asset_number: itemDetails.asset_number,
+      serial_number: itemDetails.serial_number,
+      delivered_date:
+        itemDetails.delivered_date &&
+        isValidDateTime(itemDetails.delivered_date)
+          ? new Date(itemDetails.delivered_date).toISOString() ?? null
+          : null,
+      notes: itemDetails.notes,
+    };
+  }, [itemDetails, catalogueItem]);
```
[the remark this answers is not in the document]
joshuadkitenge: This is going to be used for the handleEditItem as well. I'll add the type on it to make it clearer what it is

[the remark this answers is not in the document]
joelvdavies: Shouldn't edit use undefined instead though?

src/items/items.component.tsx
```diff
@@ -0,0 +1,82 @@
...
+        width: '100%',
+      }}
+    >
+      <Button
+        sx={{ mx: 0.5 }}
+        component={Link}
+        startIcon={
+          <Box sx={{ alignItems: 'center', display: 'flex' }}>
+            <ArrowBackIcon fontSize="small" />{' '}
+            <ArrowBackIcon fontSize="small" />
```
MatteoGuarnaccia5, line 38: image
Double arrow makes it look like this. Shouldn't it be a single arrow?

joshuadkitenge: It meant to be a double as you are going back up two levels

src/items/itemDialog.component.tsx
```diff
@@ -0,0 +1,632 @@
...
+          }}
+        >
+          <Button
+            variant="outlined"
+            sx={{ width: '50%', mx: 1 }}
+            onClick={handleClose}
+          >
+            Cancel
+          </Button>
+          <Button
```
MatteoGuarnaccia5, line 614: add logic so save button gets disabled when there is an error in the dialog. see #189

MatteoGuarnaccia5 (commented):
Just a couple of things from me (see above)

MatteoGuarnaccia5 (approved):
LGTM

joelvdavies (approved):

Issue #226, opened 2024-01-10
Create a Common folder


joshuadkitenge:
- common components
- common functions

joelvdavies:
Also https://github.com/ral-facilities/inventory-management-system/pull/224#discussion_r1448824713

joshuadkitenge:
Move isValidDateTime #240

PR #293, opened 2024-02-01
allowed value list in the dialogs #292

### LINKED ISSUE #278: edit item propeties bug
- send back null if the value is undefined 
- Items 
- catalogue items 
- edit the mock data to return none 
- check the landing pages

### LINKED ISSUE #292: Implement the allowed values list in the items and catalogue items dialogues

joshuadkitenge:
## Description

When the catalogue item property has allowed list of values it renders a select component with the options  

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes #292

### DIFF
--- a/src/catalogue/items/catalogueItemsDialog.component.test.tsx
+++ b/src/catalogue/items/catalogueItemsDialog.component.test.tsx
@@ -212,6 +212,62 @@ describe('Catalogue Items Dialog', () => {
     });
   }, 10000);
 
+  it('adds a catalogue item where the catalogue item property has an allowed list of values', async () => {
+    props = {
+      ...props,
+      parentInfo: getCatalogueCategoryById('12'),
+    };
+
+    createView();
+
+    await modifyValues({
+      costGbp: '1200',
+      costToReworkGbp: '400',
+      daysToReplace: '20',
+      daysToRework: '2',
+      description: '',
+      drawingLink: 'https://example.com',
+      drawingNumber: 'mk4324',
+      itemModelNumber: 'mk4324',
+      name: 'test',
+      manufacturer: 'Man{arrowdown}{enter}',
+    });
+
+    await fireEvent.change(
+      screen.getByLabelText('Ultimate Pressure (millibar) *'),
+      {
+        target: { value: '10' },
+      }
+    );
+
+    await fireEvent.mouseDown(screen.getByLabelText('Pumping Speed *'));
+    await fireEvent.click(within(screen.getByRole('listbox')).getByText('400'));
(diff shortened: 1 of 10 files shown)


src/items/itemDialog.component.tsx
```diff
@@ -649,6 +649,49 @@ function ItemDialog(props: ItemDialogProps) {
                                 </FormHelperText>
```
joelvdavies, line 649: The tooltips go oval shaped when there are errors
image

src/items/itemDialog.component.test.tsx
```diff
@@ -200,6 +200,45 @@ describe('ItemDialog', () => {
       });
     });
 
+    it('adds an item where the item property has an allowed list of values', async () => {
```
joelvdavies, line 203: Could this also have a string allowed_values list instead of only numbers, there are a few others as well.

src/catalogue/items/catalogueItemsDialog.component.test.tsx
```diff
@@ -220,6 +220,103 @@ describe('Catalogue Items Dialog', () => {
     });
   }, 10000);
 
+  it('adds a catalogue item where the catalogue item property has an allowed list of values', async () => {
```
joelvdavies, line 223: Could this also have a string allowed_values list instead of only numbers.

src/items/itemDialog.component.tsx
```diff
@@ -649,6 +649,49 @@ function ItemDialog(props: ItemDialogProps) {
...
                             </FormControl>
+                          ) : property.allowed_values ? (
+                            <FormControl fullWidth>
+                              <InputLabel
+                                required={property.mandatory ?? false}
+                                error={propertyErrors[index]}
+                                id={`catalogue-item-property-${property.name.replace(
+                                  /\s+/g,
+                                  '-'
+                                )}`}
```
joelvdavies, line 660: Could this just go into utils as a sanitise_name/remove_whitespace or something? - its used in multiple places

joshuadkitenge: #226

joelvdavies: Its a very small change in this case and I feel like if we come up with a case this doesn't work it would be annoying to change even though we can just put it in a common place now.

joshuadkitenge: The plan is to do each of the general change in separate prs anyways.

joelvdavies (approved):

PR #322, opened 2024-02-14
Start move and copy from the same place everywhere #318

### LINKED ISSUE #318: Move and copy catalogue Items starts in directory above
I think it should really start in the same directory e.g. if copying to the current directory, navigation shouldn't be necessary. This is what systems & the obsolete dialogue does.

Other small things
- [ ] Clicking cancel resets the current selection
- [ ] Systems move dialog doesn't reset location when cancelled
- [ ] Can reduce catalogue category re-renders when opening and closing the move/copy dialogues by moving buttons into their own components & doing navigation in the dialogue itself
- [ ] Replace deprecated uses of `toBeCalledWith` -> `toHaveBeenCalledWith`, `toBeCalled` -> `toHaveBeenCalled`, `toBeCalledTimes` -> `toHaveBeenCalledTimes`

joelvdavies:
## Description

See #318 for a full list of changes.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

Closes #318


VKTB (changes requested):
The state does not reset when I try to move a catalogue item, click cancel and then Move To again - it starts from where the catalogue category I was in before. The other cases work well.

VKTB (approved):

Issue #152, opened 2023-12-05
Scigateway intergration


joshuadkitenge:
- [x] Create a homepage for IMS (non-mvp)
- [x] Check that the css styling works as expected  
- [x] #355
- [x] The error and success handling
- [x] #357

PR #327, opened 2024-02-20
home page #236

### LINKED ISSUE #236: HomePage

MatteoGuarnaccia5:
## Description

Created home page for IMS using template in scigateway

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

closes #236


src/view/viewTabs.component.tsx
```diff
@@ -99,7 +104,35 @@ function ViewTabs() {
...
+            greenSwirl1Image={GreenSwirl1Image}
+            greenSwirl2Image={GreenSwirl2Image}
+            decal1Image={Decal1Image}
+            decal2Image={Decal2Image}
+            decal2DarkImage={Decal2DarkImage}
+            decal2DarkHCImage={Decal2DarkHCImage}
+            facilityImage={FacilityImage}
+          />
+        }
+      />
```
joelvdavies, line 120: Could this be put into a a separate component in homePage.component, perhaps the current should be BaseHomePage or something a bit like the TranslatedHomepage in datagateway, and then this just HomePage so we don't have to import all of the images in the viewTabs component.

src/homePage/homePage.component.tsx
```diff
@@ -0,0 +1,324 @@
...
+            </Grid>
+          </Grid>
+        </Paper>
+        <Grid container spacing={2}>
+          <Grid item sm={12} md={4}>
+            <Paper sx={paperStyles} elevation={1}>
+              <Box sx={paperContentStyles}>
+                <PaperHeading variant="h4">{'Systems'}</PaperHeading>
+                <PaperDescription variant="body1">
+                  {'Navigate different systems and their sub systems.'}
```
joelvdavies, line 229:
```suggestion
                  {'Navigate through systems and their sub systems to view their component items.'}
```

src/homePage/homePage.component.tsx
```diff
@@ -0,0 +1,324 @@
...
+                    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                    color: (theme as any).colours?.homePage?.heading,
+                    marginBottom: theme.spacing(2),
+                  })}
+                >
+                  {'Catalogue'}
+                </Typography>
+                <PaperDescription variant="body1">
+                  {
+                    'Browse catalogue categories, sub categories and catalogue items.'
```
joelvdavies, line 186:
```suggestion
                    'Browse catalogue categories, sub categories and catalogue items to locate specific items or add new ones.'
```

src/homePage/homePage.component.tsx
```diff
@@ -0,0 +1,324 @@
...
+                  }
+                </PaperDescription>
+                <Box marginTop="16px">
+                  <Button
+                    color="primary"
+                    variant="contained"
+                    component={Link}
+                    to={'/catalogue/*'}
```
joelvdavies, line 194:
```suggestion
                    to={'/catalogue'}
```
This doesn't work currently.

src/homePage/homePage.component.tsx
```diff
@@ -0,0 +1,324 @@
...
+                <PaperHeading variant="h4">{'Systems'}</PaperHeading>
+                <PaperDescription variant="body1">
+                  {'Navigate different systems and their sub systems.'}
+                </PaperDescription>
+                <Box marginTop="auto">
+                  <Button
+                    color="primary"
+                    variant="contained"
+                    component={Link}
+                    to={'/systems/*'}
```
joelvdavies, line 236:
```suggestion
                    to={'/systems'}
```

joelvdavies (changes requested):
A few suggestions. I was going to raise the fact the buttons all say Explore, although I can't really come up with anything less repetitive 😄. I think an image of EPAC itself may fit nicely as the top image but we can raise this with them later.

src/homePage/homePage.component.tsx
joelvdavies: Could you  also add some e2e tests for the homepage, just for the links e.g. https://github.com/ral-facilities/datagateway/blob/develop/packages/datagateway-dataview/cypress/e2e/homePage.cy.ts

src/images/facility.jpg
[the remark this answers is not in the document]
MatteoGuarnaccia5: To substitute which image? The top banner or the image in the middle ?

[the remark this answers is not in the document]
joelvdavies: Image in the middle, it seems about the right aspect ratio.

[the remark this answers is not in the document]
MatteoGuarnaccia5: image

I can crop the photo so we get the top of the building, but this does not look that nice to me

src/homePage/homePage.component.tsx
```diff
@@ -0,0 +1,326 @@
...
+    <HomePage
+      logo={DGLogo}
+      backgroundImage={BackgroundImage}
+      greenSwirl1Image={GreenSwirl1Image}
+      greenSwirl2Image={GreenSwirl2Image}
+      decal1Image={Decal1Image}
+      decal2Image={Decal2Image}
+      decal2DarkImage={Decal2DarkImage}
+      decal2DarkHCImage={Decal2DarkHCImage}
+      facilityImage={FacilityImage}
```
joelvdavies, line 323: I tried testing whether we needed `pluginHost` here like in https://github.com/ral-facilities/datagateway/blob/develop/packages/datagateway-dataview/src/page/translatedHomePage.component.tsx and originally it seemed we didn't when I used the production docker compose file, but when I tested it locally in development it seems we actually do as replacing one of the images and running it in SciGateway gives
image, but adding `"http://localhost:5001" +` to the image works. So I think we are going to need the same here afterall.

Also the button colours are wrong in dark mode in this image due to https://github.com/ral-facilities/inventory-management-system/blob/a15cde2807168b692917ac8628e6f08dd2c45419/src/imsThemeProvider.component.tsx#L29-L37. @joshuadkitenge Why was this done?

joelvdavies: Spoke to Joshua about the button colours, its like this as it was changed for OperationsGateway - I will make a note on #152 to address it later as I think it will require a SciGateway change.

src/homePage/homePage.component.tsx
```diff
@@ -0,0 +1,332 @@
...
+          </Grid>
+        </Grid>
+      </Box>
+    </div>
+  );
+};
+
+export interface HomePageProps {
+  pluginHost: String;
+}
```
joelvdavies, line 313: Can you obtain `pluginHost` from the settings file inside HomePage instead? It needs to be configurable for the production instance. In datagateway we were using the old style of redux (https://github.com/ral-facilities/datagateway/blob/develop/packages/datagateway-dataview/src/page/translatedHomePage.component.tsx#L45-L50), but I don't think redux should be necessary here.

MatteoGuarnaccia5: I'm struggling to find a way to get it, and get it to work as well

joelvdavies: Ah I see, yes its async. In that case I get the feeling we need to use redux in this case as well. I am having a chat with Joshua about this after lunch.

joelvdavies: We can avoid it for now using something like this
```
    // TODO: Replace this with redux if it ends up being used elsewhere
    const [pluginHost, setPluginHost] = useState<string>('');
    useEffect(() => {
      const loadPluginHost = async () =>
        setPluginHost((await settings)?.pluginHost || '');
      loadPluginHost();
    }, []);
```
and then we don't need the prop anymore either.

joelvdavies (approved):

PR #348, opened 2024-02-27
Migrate from CRA to Vite #346

### LINKED ISSUE #346: Migrate CRA to Vite
Create react app has not been maintained since April 2022.

joelvdavies:
## Description

Migrates from create react app to Vite. `yarn install` before reviewing.

## Notes

Originally followed https://www.freecodecamp.org/news/how-to-migrate-from-create-react-app-to-vite/, while also referring to a new  example created with `yarn create vite`. Used `yarn add vite @vitejs/plugin-react` when adding packages, as saw no reason to use any of the plugins yet and didn't use SWC as its newer and I believe we were already using babel anyway.

Main changes to note:
- `yarn start` -> `yarn dev`
- `yarn serve:build` -> `yarn preview`
- Instances of `process.env.REACT_` replaced with `import.meta.env.VITE_`
- MSW wouldn't load with Vite - it just hung on start - fixed by MSW v2 so upgraded that at the same time
- Images inside src don't get served like they used to so need to be in the public folder and imported with root paths e.g. `../images/facility.jpg` -> `/images/facility.jpg` (`pluginHost` is still required)
- Vite says cjs is deprecated, so package.json needs `"type": "module"` (Main thing  is `require` should be replaced with `await import`)
- Vitest requires global mode to be like jest - particualry to avoid needing cleanup from react testing library (see https://vitest.dev/guide/migration.html#globals-as-a-default and https://vitest.dev/config/#globals)
- Renamed setupTests.tsx to testUtils.tsx (Removing MSW to a separate new setupTests.tsx imported in the vitest config - to avoid repeated execution during imports)
- `jest.setTimeout(milliseconds)` is replaced by `vi.setConfig({ testTimeout: milliseconds })`
- Replace `jest.requireActual` with `await vi.importActual`
- Vite still had process.env in the build for SciGateway - applied fix from https://stackoverflow.com/questions/74389961/how-to-remove-process-env-node-env-from-vites-production-bundle-and-delete-unus
- It is not possible to apply the rollup config while in dev mode at the moment, so to run in dev mode through SciGateway use `preview:build:dev` which does `yarn build --watch`
- I have used `npx vite-bundle-visualizer` to check the bundle after build to ensure msw is excluded
image
- @testing-library/jest-dom is still required (it supports vitest as described https://testing-library.com/docs/svelte-testing-library/setup/)
- Removed `cypress-failed-log` as 2 years old and I don't believe it does anything anymore or otherwise the import is broken as it cannot use require

Notes for running:
- I had an issue with port forwarding through vscode ssh, meaning that yarn dev caused errors like https://github.com/microsoft/vscode-remote-release/issues/9548 - Fixed with vscode 1.87.1, or by disabling `useExecServer`
- I recommend using the vitest plugin on vscode if applicable (disable/remove any existing jest one first)

## Future tasks
- eslint-config-react-app should also be replaced (with eslint-plugin-react, eslint-plugin-testing-library, eslint-plugin-react-hooks, eslint-plugin-vitest or equivalent) - see https://github.com/vitejs/vite/tree/main/packages/create-vite/template-react-ts - done in #372 

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Ensure linting works locally (in src and unit & e2e tests)
- [ ] Test dev, preview:build, and preview:build:dev locally

## Agile board tracking

Closes #346


VKTB (approved):
I tested the production Docker image and it seems to be building and working fine.

louise-davies (dismissed):
LGTM - as I said there's a minor issue I had with the "build watch" mode where sometimes the build would have a syntax error, but re-saving a file to trigger a rebuild normally fixed it, plus it seems like it's a problem with the build watch mode which is a developer convenience and not a production concern.

joshuadkitenge (approved):
LGTM

PR #349, opened 2024-02-27
Make pluginHost setting configurable in production Docker Image

VKTB:
## Description
Makes the `pluginHost` setting configurable in the production Docker image. Also updates the `httpd` Docker image.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Check that the returned IMS settings JSON file has the correct `pluginHost` value


joelvdavies (approved):
Tested on the production compose where it worked correctly and could be overridden.

Issue #287, opened 2024-01-30
Use react-hook-form library for the dialogs


joshuadkitenge:
https://www.react-hook-form.com/

[schema (zod)](https://react-hook-form.com/docs/useform#resolver)

### Manufacturers
- [x] #664 
- [x] #674
- [x] #663 

### Units
- [x] #681
- [x] #683
- [x] #685

### Usage Status
- [x] #687
- [x] #691
- [x] #690

### systems
- [x] #694
- [x] #695
- [x] #696

### Catalogue Categories
- [x] #710
- [x] #711
- [x] #720

### Catalogue Items
- [x] #936
- [x] #937
- [x] #946

### Items
- [x] #963
- [x] #964
- [x] #978

PR #359, opened 2024-02-29
React hook for manufacturer #287

joshuadkitenge:
## Description

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

connect to {issue number}


src/manufacturer/manufacturerDialog.component.tsx
joelvdavies, line 1: When getting a duplicate name error, changing the name doesn't allow save to be clicked.

src/manufacturer/manufacturerDialog.component.tsx
```diff
@@ -10,79 +14,51 @@ import {
...
-  ManufacturerDetails,
+  Manufacturer,
 } from '../app.types';
-import { useAddManufacturer, useEditManufacturer } from '../api/manufacturer';
 import { AxiosError } from 'axios';
+import { useAddManufacturer, useEditManufacturer } from '../api/manufacturer';
 import handleIMS_APIError from '../handleIMS_APIError';
-import { trimStringValues } from '../utils';
+
+const manufacturerSchema = z.object({
```
joelvdavies, line 27: The values sent are all empty strings except URL which is undefined. Ideally they would all be undefined or null when not given as the database schema is specifically either null or a value.
image

src/manufacturer/manufacturerDialog.component.tsx
```diff
@@ -10,79 +14,51 @@ import {
...
+  name: z.string().trim().min(1, { message: 'Please enter a name.' }),
+  url: z
+    .string()
+    .url({ message: 'Please enter a valid URL' })
```
joelvdavies, line 31: When editing and removing the URL, it gives
image

I would personally do something like
const transformEmptyStringToNull = (value: string) =>
  value === '' ? null : value;

const ZodStringOrNull = z.string().trim().transform(transformEmptyStringToNull);
const ZodUrlOrNull = ZodStringOrNull.transform((value) =>
  z
    .string()
    .url({ message: 'Please enter a valid URL' })
    .nullable()
    .parse(value)
);

So they can all be reused. This for the url also means it does the trim/null comparison first so a url of "    " would also be properly accounted for.

joelvdavies (commented):
A few things I noticed

src/manufacturer/manufacturerDialog.component.tsx
```diff
@@ -92,419 +68,269 @@ function ManufacturerDialog(props: ManufacturerDialogProps) {
...
+      ? handleAddManufacturer(data)
+      : handleEditManufacturer(data);
+  };
 
   return (
-    <Dialog open={open} onClose={handleClose} maxWidth="lg" fullWidth>
-      <DialogTitle>{`${
-        type === 'create' ? 'Add' : 'Edit'
-      } Manufacturer`}</DialogTitle>
+    <Dialog open={open} onClose={onClose} maxWidth="lg" fullWidth>
```
joelvdavies, line 264:
```suggestion
    <Dialog open={open} onClose={handleClose} maxWidth="lg" fullWidth>
```

src/manufacturer/manufacturerDialog.component.tsx
```diff
@@ -525,25 +351,20 @@ function ManufacturerDialog(props: ManufacturerDialogProps) {
           <Button
             variant="outlined"
             sx={{ width: '50%', mx: 1 }}
-            onClick={handleClose}
+            onClick={onClose}
```
joelvdavies, line 354:
```suggestion
            onClick={handleClose}
```

src/manufacturer/manufacturerDialog.component.tsx
joshuadkitenge: The issue with this is error message. The error message are tailored to each text field

joshuadkitenge:
#663

PR #368, opened 2024-03-06
Migration to RouterProvider #235

### LINKED ISSUE #235: use RouterProvider react  router v6
https://reactrouter.com/en/main/upgrading/v6-data

MatteoGuarnaccia5:
## Description

Migrated routes to use RouterProvider v6. refactored tests to use new logic.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

closes #235


src/view/viewTabs.component.tsx
```diff
@@ -135,12 +110,12 @@ function ViewTabs() {
...
           >
             <TabPanel value={value} label={value}>
-              {routing}
+              <Outlet />
             </TabPanel>
           </Box>
         </>
       ) : (
-        routing
+        <></>
```
joshuadkitenge, line 118:
```suggestion
        <Outlet />
```

src/view/viewTabs.component.tsx
```diff
@@ -3,19 +3,12 @@ import Tab from '@mui/material/Tab';
...
-import Items from '../items/items.component';
-import ItemsLandingPage from '../items/itemsLandingPage.component';
-import Manufacturer from '../manufacturer/manufacturer.component';
-import ManufacturerLandingPage from '../manufacturer/manufacturerLandingPage.component';
-import Systems from '../systems/systems.component';
 import { getSciGatewayPageHeightCalc, isRunningInDevelopment } from '../utils';
-import { HomePage } from '../homePage/homePage.component';
 
 export const paths = {
+  any: '*',
```
joelvdavies, line 11: Could this also be moved to `App.tsx` and then be used inside the router itself.

src/App.tsx
```diff
@@ -69,28 +77,60 @@ const App: React.FunctionComponent = () => {
...
   }, []);
 
+  return <div className="Root"></div>;
+};
+
+const router = createBrowserRouter([
+  {
+    path: '/*',
+    Component: Layout,
+    children: [
```
joelvdavies, line 87: Should this even have a path? The migration guide FAQ suggests it should be pathless.

src/App.tsx
```diff
@@ -47,7 +55,7 @@ const queryClient = new QueryClient({
   },
 });
 
-const App: React.FunctionComponent = () => {
+const Root: React.FunctionComponent = () => {
```
joelvdavies, line 58: I don't think `Root` is actually needed any more after the migration - I would assume Layout would deal with rendering something. The contents of this though should probably be in `Layout` as right now none of it does anything.

joelvdavies (changes requested):
Haven't looked at the test changes yet, but I did notice an issue with the plugin refresh not doing anything.

Also just to raise, I saw https://reactrouter.com/en/main/start/tutorial#handling-not-found-errors allows us to have 404 pages for invalid ids, and more general error pages we could look at implementing.

src/view/breadcrumbs.component.test.tsx
```diff
@@ -1,15 +1,19 @@
...
 describe('Channel Breadcrumbs', () => {
   const onChangeNode = vi.fn();
   const onChangeNavigateHome = vi.fn();
   let props: BreadcrumbsProps;
 
   const createView = () => {
-    return renderComponentWithBrowserRouter(<Breadcrumbs {...props} />);
+    return renderComponentWithRouterProvider(
+      <Breadcrumbs {...props} />,
+      paths.root
```
joelvdavies, line 15:
```suggestion
```
I don't think this does anything in the most recent commit.

src/testUtils.tsx
```diff
@@ -30,56 +35,34 @@ interface ExtendedRenderOptions extends Omit<RenderOptions, 'queries'> {
   queryClient?: QueryClient;
 }
 
-export function renderComponentWithBrowserRouter(
+export function renderComponentWithRouterProvider(
   ui: React.ReactElement,
+  urlPathKey?: keyof typeof paths,
+  pathName?: string,
```
joelvdavies, line 41:
```suggestion
  initialEntry?: string,
```
I think this makes it a bit clearer.

src/view/viewTabs.component.test.tsx
```diff
@@ -20,13 +20,7 @@ describe('View Tabs', () => {
     expect(viewTabs.getByRole('tab', { selected: true })).toHaveTextContent(
       'Catalogue'
     );
-    const catalogueHomeButton = await screen.findByRole('button', {
-      name: 'navigate to catalogue home',
-    });
-    expect(catalogueHomeButton).toBeInTheDocument();
-    await waitFor(() => {
-      expect(screen.getByText('Beam Characterization')).toBeInTheDocument();
-    });
```
joelvdavies, line 29: Why was this removed? It ensures the routing renders the appropriate component. Same with the other removals below.

joelvdavies (commented):
A couple comments from the tests.

src/view/viewTabs.component.tsx
joelvdavies: @MatteoGuarnaccia5 I think you missed out this.

MatteoGuarnaccia5: I did add it to `App.tsx`, this any was just for `vewTabs.component.test.tsx` as it doesn't work with any other of the paths

joelvdavies: Ah sorry, I meant the whole `paths` object so they are imported from App.tsx and used inside App.tsx inside the router for the paths.

joelvdavies (approved):
Tested locally in dev, locally in SciGateway with and without auth (checking token expiry works) and with the dev docker compose used on the dev machine.

joshuadkitenge (approved):

PR #376, opened 2024-03-11
Rhf with zod cc #287

joshuadkitenge:
## Description

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

connect to {issue number}


joelvdavies:
Not reviewing properly just yet, but noticed when adding a number allowed values list with some invalid text, and then switching the type from text to number it doesn't reset the state so save can't be clicked.

PR #407, opened 2024-03-18
allow the user to select None when a allowed list of values is used #391

### LINKED ISSUE #391: allow the user to select None when a allowed list of values is used

joshuadkitenge:
## Description

allow the user to select None when a allowed list of values is used

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #391

### DIFF
--- a/src/catalogue/items/catalogueItemsDialog.component.test.tsx
+++ b/src/catalogue/items/catalogueItemsDialog.component.test.tsx
@@ -703,6 +703,43 @@ describe('Catalogue Items Dialog', () => {
       });
     }, 10000);
 
+    it('edits a catalogue item where the catalogue item property has an allowed list of values (Optional)', async () => {
+      props = {
+        ...props,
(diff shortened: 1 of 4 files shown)


joelvdavies (changes requested):
I think none should only be an option if the field is not mandatory.

joshuadkitenge:
> I think none should only be an option if the field is not mandatory.

We have error handling for that case. It makes sense that the drop down to be consistent

MatteoGuarnaccia5 (approved):
LGTM

joelvdavies:
> > I think none should only be an option if the field is not mandatory.
> 
> We have error handling for that case. It makes sense that the drop down to be consistent.
> 
> also the default value should always also be in the dropdown (default value : '' (add))

The value should need to be there e.g. https://mui.com/material-ui/react-select/ provided it has the right type. But I see what you mean.

joelvdavies (approved):

PR #412, opened 2024-03-18
Fix spacing items details panels #394

### LINKED ISSUE #394: Details panel has properties too close together (catalogue item and items details)
Compare
image
with
image
The spacing should really be consistent between the details, properties and manufacturers tabs.

Also should remove the Notes heading and either have the actual notes or `None` displayed in the tab panel.

joelvdavies:
## Description

Adds column spacing to properties on catalogue item & item details panels. Ideally we would use 100% width on the grid container so the grid could use `space-between` or something but unfortunately MRT renders MUI collapse elements above with a width that can't be modified, and using the example given on MRT fails when columnResizing is `true` currently.

This PR also removes some unnecessary Grid's and moves Notes from the `Details` tab to a separate `Notes` tab on catalogue item details panels to match the items one.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

Closes #394

### DIFF
--- a/src/catalogue/items/CatalogueItemsDetailsPanel.component.test.tsx
+++ b/src/catalogue/items/CatalogueItemsDetailsPanel.component.test.tsx
@@ -63,4 +63,20 @@ describe('Catalogue Items details panel', () => {
 
     expect(view.asFragment()).toMatchSnapshot();
   });
+
+  it('renders notes panel correctly', async () => {
+    const view = createView();
+    await user.click(screen.getByText('Notes'));
+
+    expect(view.asFragment()).toMatchSnapshot();
+  });
+
+  it('renders details panel correctly (when there are no Notes)', async () => {
+    props.catalogueCategoryData = getCatalogueCategoryById('4');
+    props.catalogueItemIdData = getCatalogueItemById('33');
+
+    const view = createView();
+
+    expect(view.asFragment()).toMatchSnapshot();
+  });
 });
--- a/src/catalogue/items/CatalogueItemsDetailsPanel.component.tsx
+++ b/src/catalogue/items/CatalogueItemsDetailsPanel.component.tsx
@@ -160,21 +160,17 @@ function CatalogueItemsDetailsPanel(props: CatalogueItemsDetailsPanelProps) {
               </Grid>
 
               <Grid item xs={12} sm={6} key={9}>
-                <Grid item xs={12}>
-                  <Typography color="text.primary">Last Modified</Typography>
-                  <Typography color="text.secondary">
-                    {formatDateTimeStrings(catalogueItemIdData.modified_time)}
-                  </Typography>
-                </Grid>
+                <Typography color="text.primary">Last Modified</Typography>
+                <Typography color="text.secondary">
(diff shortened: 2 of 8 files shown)


joshuadkitenge (changes requested):
It looks like the gap is getting bigger when you increase the amount properties.
  
image
image


you can recreate this using the mock data (cameras and wavefront sensors)

joelvdavies:
> It looks like the gap is getting bigger when you increase the amount properties.
> 
> image image
> 
> you can recreate this using the mock data (cameras and wavefront sensors)

Yeah this what I mentioned yesterday. It does this for all tabs and I have no idea how to fix it 🙃 I will try though.

joshuadkitenge:
A simpler solution is just to force it to only fill one of the two columns by changing xs={6} instead of xs={9} 

```
      <TabPanel value={tabValue} index={1}>
          <Grid item container spacing={0}>
            {catalogueItemIdData.properties &&
              catalogueItemIdData.properties.map((property, index) => (
                <Grid item xs={6} sm={6} key={index}>
                  <Typography color="text.primary">{`${property.name} ${
                    property.unit ? `(${property.unit})` : ''
                  }`}</Typography>
                  <Typography color="text.secondary">
                    {property.value !== null ? String(property.value) : 'None'}
                  </Typography>
                </Grid>
              ))}
          </Grid>
        </TabPanel>
```

joelvdavies:
> A simpler solution is just to force it to only fill one of the two columns by changing xs={6} instead of xs={12}
> 
> ```
>       <TabPanel value={tabValue} index={1}>
>           <Grid item container spacing={0}>
>             {catalogueItemIdData.properties &&
>               catalogueItemIdData.properties.map((property, index) => (
>                 <Grid item xs={6} sm={6} key={index}>
>                   <Typography color="text.primary">{`${property.name} ${
>                     property.unit ? `(${property.unit})` : ''
>                   }`}</Typography>
>                   <Typography color="text.secondary">
>                     {property.value !== null ? String(property.value) : 'None'}
>                   </Typography>
>                 </Grid>
>               ))}
>           </Grid>
>         </TabPanel>
> ```

Unfortunately that doesn't fix the original issue:
image

joshuadkitenge:
> > A simpler solution is just to force it to only fill one of the two columns by changing xs={6} instead of xs={12}
> > ```
> >       <TabPanel value={tabValue} index={1}>
> >           <Grid item container spacing={0}>
> >             {catalogueItemIdData.properties &&
> >               catalogueItemIdData.properties.map((property, index) => (
> >                 <Grid item xs={6} sm={6} key={index}>
> >                   <Typography color="text.primary">{`${property.name} ${
> >                     property.unit ? `(${property.unit})` : ''
> >                   }`}</Typography>
> >                   <Typography color="text.secondary">
> >                     {property.value !== null ? String(property.value) : 'None'}
> >                   </Typography>
> >                 </Grid>
> >               ))}
> >           </Grid>
> >         </TabPanel>
> > ```
> 
> Unfortunately that doesn't fix the original issue: image

weren't the numbers overlapping before ?

joshuadkitenge (approved):
LGTM

PR #435, opened 2024-03-20
Resolve authenticated request immediately if not logged in #399

### LINKED ISSUE #399: Scigateway refresh issue
refresh issue

joelvdavies:
## Description

Resolves requests immediately if the user is not logged in and is trying to access a protected resource, rather than before which attempted to refresh the token. Also hides the error message when this occurs (as well as any 403 as it is assumed this should be handled by SciGateway like in DataGateway).

I also noticed that the status check was wrong, meaning the error message for a 500 error would not be used so I fixed that here.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

Closes #399

### DIFF
--- a/src/api/api.tsx
+++ b/src/api/api.tsx
@@ -45,11 +45,13 @@ imsApi.interceptors.response.use(
       : error.message;
 
     // Check if the token is invalid and needs refreshing
-    // only allow a request to be retried once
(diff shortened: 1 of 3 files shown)


joshuadkitenge (approved):
LGTM

VKTB (approved):

PR #436, opened 2024-03-21
Pin docker images to SHA

VKTB:
## Description
This PR pins the base docker images in the Dockerfiles to commit SHAs so that Renovate (or Dependabot if Renovate doesn't work) can create PRs each time there is a change in the upstream base images that the base images are set to use as new SHAs are generated on every upstream change. Not sure if Renovate will also update the OS versions but if it doesn't then we will have to do the Alpine version updates manually for now.

Please note that I have pinned it to an older SHA to verify that Renovate (or Dependabot if Renovate doesn't work) will create a PR.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -82,7 +82,7 @@ jobs:
           MONGO_INITDB_ROOT_PASSWORD: example
 
       ims-api:
-        image: harbor.stfc.ac.uk/inventory-management-system/ims-api:develop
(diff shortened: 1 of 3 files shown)


joelvdavies (approved):

VKTB:
Renovate created PRs to update the SHAs https://github.com/ral-facilities/inventory-management-system/pull/443/files and https://github.com/ral-facilities/inventory-management-system/pull/444

PR #449, opened 2024-03-25
prevent dialogs closing on background clicks #442

### LINKED ISSUE #442: Prevent closing dialogues by clicking off them
To avoid an accidental click from loosing data. (Change handleClose to onClose in the dialogue component, see https://mui.com/material-ui/api/dialog/, disable only for backdropClick))

MatteoGuarnaccia5:
## Description

Added logic on `onClose` prop in dialog components to prevent them closing on background clicks

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

closes #442


joshuadkitenge (changes requested):
can you on check the close functions in all of the dialogs.  if there is a handleClose the dialog should be using that for it onClose

joelvdavies (changes requested):
Pressing escape doesn't work. Also had these warnings when doing a yarn build:
image

src/catalogue/category/catalogueCategoryDialog.component.tsx
```diff
@@ -492,7 +492,9 @@ const CatalogueCategoryDialog = React.memo(
     return (
       <Dialog
         open={open}
-        onClose={(reason) => reason == 'backdropClick' ?? handleClose}
+        onClose={(event, reason) =>
+          event && reason == 'backdropClick' ? null : handleClose()
+        }
```
joelvdavies, line 497:
```suggestion
        onClose={(_event, reason) =>
          reason !== 'backdropClick' && handleClose()
        }
```
This should also work without needing to include event which should never be empty here anyway. It also means if any other reasons are added in MUI in the future the logic will still work. Same for the others.

src/systems/systemItemsDialog.component.tsx
```diff
@@ -79,7 +79,9 @@ const SystemItemsDialog = React.memo((props: SystemItemsDialogProps) => {
   return (
     <Dialog
       open={open}
-      onClose={(reason) => reason == 'backdropClick' ?? onClose}
+      onClose={(event, reason) =>
+        event && reason == 'backdropClick' ? null : onClose()
+      }
```
joelvdavies, line 84:
```suggestion
      onClose={(_event, reason) =>
        reason !== 'backdropClick' && handleClose()
      }
```
Handle close is present above.

MatteoGuarnaccia5: It isn't used for the cancel button, shall I change the button to use handleClose ?

joelvdavies: Yes

joshuadkitenge (changes requested):
Can you add unit tests please for ESCAPE key down and backdrop key down

joelvdavies (dismissed):
LGTM

joshuadkitenge (changes requested):
I had a chat with Louise, and I think it would be better if the ESC keydown and the backdrop click acted the same way. This consistency is important for both keyboard and mouse users. The only way to close the dialog should be through the cancel button.

src/catalogue/category/catalogueCategoryDialog.component.tsx
```diff
@@ -490,7 +490,15 @@ const CatalogueCategoryDialog = React.memo(
     ]);
 
     return (
-      <Dialog open={open} onClose={handleClose} maxWidth="lg" fullWidth>
+      <Dialog
+        open={open}
+        onClose={(_event: object, reason: string) =>
```
joshuadkitenge, line 495: Could we explore alternative implementations? Much of the existing logic appears redundant now.

joshuadkitenge (changes requested):

joelvdavies (approved):

joshuadkitenge (approved):

PR #452, opened 2024-03-25
removed item ID and moved serial num column #440

### LINKED ISSUE #440: Remove item id from items table and use serial number as first column instead
When there is no serial number use some text e.g. No serial number and have the link to the landing page be that.
Also remove grouping - Doesn't make sense as they should be different (or empty).

MatteoGuarnaccia5:
## Description

- removed item ID from table
- Moved Serial number to be first column and made them be links to landing page
- Removed grouping

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

closes #440


src/items/itemsTable.component.tsx
```diff
@@ -77,14 +77,13 @@ export function ItemsTable(props: ItemTableProps) {
...
+        accessorFn: (row) => row.serial_number,
+        id: 'serial_number',
         size: 250,
-        accessorFn: (row) => row.id,
-        id: 'id',
-        enableGrouping: false,
         Cell: ({ row }) => (
           <MuiLink underline="hover" component={Link} to={row.original.id}>
-            {row.original.id}
+            {row.original.serial_number ?? 'No Serial number'}
```
joelvdavies, line 86:
```suggestion
            {row.original.serial_number ?? 'No serial number'}
```

src/items/itemsTable.component.tsx
```diff
@@ -301,7 +291,6 @@ export function ItemsTable(props: ItemTableProps) {
     enableRowVirtualization: false,
     enableFullScreenToggle: false,
     enableColumnVirtualization: dense ? false : true,
-    enableGrouping: !dense,
```
joelvdavies, line 304: Sorry when I said remove grouoping I should have said only for the serial number 😄 please undo this change, and ensure `enableGrouping: false` is added to the serial number column instead. Also undo the changes for created and modified as the grouping will never make sense due to the seconds being included. Asset number is probably ok though.

joelvdavies (changes requested):

src/items/itemsTable.component.tsx
MatteoGuarnaccia5: Ohh my bad  😄

joelvdavies:
Also just realised the filtering doesn't respect the `No serial number` when I think ideally it should.

joelvdavies:
Could you also make the landing page breadcrumbs display the serial number instead.
image
The SystemItemsTable also needs the item ID switching with the serial number in the same way, displaying them as `No serial number` if needed.

joshuadkitenge (changes requested):
Can you make the same changes to the landing pages and the details panels for items for the serial number

joelvdavies (approved):
LGTM

joshuadkitenge (approved):

Issue #454, opened 2024-03-25
Preserving all page states  


joshuadkitenge:
- Testing the limitation of storing the table state in the url. (sorting, filtering ,pagination, page number, global filter and visible columns)

- catalogue categories only need page and pagination

- exploring other ways  and combinations e.g (local storage).

joelvdavies:
### Closes with #464

PR #464, opened 2024-03-28
Preserve page states #454

joelvdavies:
## Description

Adds a custom hook for preserving the following table states in the URL:

- Column filters
- Sorting
- Column visibility
- Global filter
- Column order
- Grouping
- Pagination state

The state is compressed using `lz-string`, leading to a reduction of about 42% of its size. I tested url lengths in edge, chrome and firefox and found that the length is unlimited unless sharing it, at which point our current settings for nginx leads to a limit of 8192 characters. This should not be hit unless a really long filter is entered.

The same hook is used for the catalogue category card view, but only using the pagination functionality.

### Main areas that need testing for the interaction with MRT
- Showing and hiding all columns at once - This is already tested by the  column virtualisation tests (and is how I found the issue in the first place)
- Pagination on page load and its first update - Snapshot should cover the page load, first update needs separate test
- Column order on page load and its first update
- Grouping also changes column order - cannot be tested in a unit test as requires drag functionality which is not supported by jsdom and could not find a way to get it to work in cypress either (seems to be related to MRT's implementation)

### Other notes

- Modified hooksWrapperWithProviders to include the router provider for the new custom hook to function in tests (and allowed the router to be returned for testing urls when using `renderComponentWithRouterProvider` - unfortunately this cannot be done for hooks due to the need of using `renderHook` with a wrapper)
- We may wish to have a further issue to preserve pageSize across navigation e.g. systems/catalogue card view

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Test editing the url params to be nonsense

## Agile board tracking

Relates to #454


joshuadkitenge:
table bug ds
@joelvdavies

joshuadkitenge (changes requested):
This works really well. I just found one bug

joshuadkitenge (approved):

PR #466, opened 2024-03-28
implemented logic to show total row count on tables #441

### LINKED ISSUE #441: Change tables to show the total number of rows
Change the pagination to include the row count e.g. https://www.material-react-table.com/docs/guides/pagination#remote-demo

MatteoGuarnaccia5:
## Description

Added text to show total rows on table and what rows are being shown in the current table state

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

closes #441


src/utils.tsx
```diff
@@ -87,3 +87,19 @@ export const formatDateTimeStrings = (
 
   return formattedDate;
 };
+
+/* returns a string with the total rows and current rows for table views */
```
joshuadkitenge, line 91: MRT should have in built function that does this, use that instead of this

MatteoGuarnaccia5: I can't find anything that does this. I used what was done in MRTs pagination on [github ](https://github.com/KevinVandy/material-react-table/blob/v2/packages/material-react-table/src/components/toolbar/MRT_TablePagination.tsx) and they don't use a function it is hardcoded like here

joshuadkitenge: The total rows should be calculated using the filters no the pagination

MatteoGuarnaccia5: image


if you apply filters it still calculates the correct number of rows.

joshuadkitenge: you shouldn't need to create a custom function for this

src/catalogue/items/catalogueItemsTable.component.tsx
```diff
@@ -772,6 +772,14 @@ const CatalogueItemsTable = (props: CatalogueItemsTableProps) => {
         </>
       );
     },
+    renderBottomToolbarCustomActions: ({ table }) => (
+      <Typography>
+        {table.getFilteredRowModel().rows.length == catalogueItemsData?.length
+          ? `Total Catalogue Items: ${catalogueItemsData.length}`
+          : `Returned ${table.getFilteredRowModel().rows.length} out of ${catalogueItemsData?.length} Catalogue Items`}
```
joelvdavies, line 779:
```suggestion
          : `Displaying ${table.getFilteredRowModel().rows.length} out of ${catalogueItemsData?.length} Catalogue Items`}
```
I think this is more user friendly.

joelvdavies (commented):
I think it might also be worth adding something like this for the subsystems table/the catalogue categories page.

src/catalogue/items/catalogueItemsTable.component.tsx
joshuadkitenge: The label 'Showing' might be misleading since it implies that the displayed number represents the current viewable catalogue items. However, it actually reflects the filtered total, which may span across multiple pages due to pagination, especially if the count is high

joelvdavies: Oh yeah sorry I see now it's slightly different to what I thought it was doing. I had assumed it was like the table pagination component, which would say `1-10 of 100` to illustrate that there are more pages to be seen. DataGrid just includes the filtering in the total so if it filtered and had one result it says `1-1 of 1`. DataGateway also does the same, changing the effective total based on the filter. I think "Returned" is alright then in this case.

src/catalogue/items/catalogueItemsTable.component.tsx
```diff
@@ -772,6 +772,14 @@ const CatalogueItemsTable = (props: CatalogueItemsTableProps) => {
         </>
       );
     },
+    renderBottomToolbarCustomActions: ({ table }) => (
+      <Typography>
```
joelvdavies, line 776: Could the padding be adjusted to align when there is grouping (/to match the other side's with the page buttons)
image

src/systems/systems.component.tsx
```diff
@@ -164,7 +164,7 @@ const columns: MRT_ColumnDef<System>[] = [
   },
 ];
 
-const MIN_SUBSYSTEMS_WIDTH = '320px';
+const MIN_SUBSYSTEMS_WIDTH = '480px';
```
joelvdavies, line 167: I think I would prefer to just have the pagination below the total, this adds a lot of empty space and is worse for small screens without changing the breakpoints further.

src/catalogue/category/catalogueCardView.component.tsx
```diff
@@ -96,6 +96,12 @@ function CatalogueCardView(props: CatalogueCardViewProps) {
         bottom={12}
         right={0}
       >
+        <Grid item position={'fixed'} left={8}>
```
joelvdavies, line 99: This causes the navigation drawer to be rendered over it rather than allowing it to move out of the way.

src/catalogue/category/catalogueCardView.component.tsx
```diff
@@ -51,25 +51,19 @@ function CatalogueCardView(props: CatalogueCardViewProps) {
     endIndex
   );
 
+  const cardViewHeight = getPageHeightCalc('100px');
```
joelvdavies, line 54: I am not sure this works correctly as if you decrease the window width the whole bottom section moves to the right:
image
setting this to 0 made it work, although it is not inline
image

MatteoGuarnaccia5: On both scigateway and just localhost300 it works fine no matter how small I make the screen. How exactly did you get it to be to the right?

joelvdavies: Just by decreasing the width of the screen. Perhaps the catalogue categories need to be filled with enough to create a scroll bar?
rowTotalBug

src/catalogue/category/catalogueCardView.component.tsx
```diff
@@ -51,25 +51,19 @@ function CatalogueCardView(props: CatalogueCardViewProps) {
     endIndex
   );
 
+  const cardViewHeight = getPageHeightCalc('0px');
```
joelvdavies, line 54: If this is 0 then it presumably isn't needed. It isn't quite fixed though, while it doesnt create a horizontal scroll bar it isn't properly aligned. Should it be in a grid with two items with xs=6?
image

joelvdavies:
It now overlaps if there are more pages/the count is larger e.g.
image

I think its worth getting it to work with `xs={12} sm={6}` on the two columns so it displays on two separate lines if its smaller e.g.
image
It didn't work properly with just that change as I also had to set the card view height to 0px again. It also likely needs usage of a breakpoint to do this to change the justifyContent in the grid for the pagingation e.g.
```
sx={(theme) => ({
    [theme.breakpoints.up('sm')]: { justifyContent: 'flex-end' },
})}
```

joelvdavies:
> It now overlaps if there are more pages/the count is larger e.g. image
> 
> I think its worth getting it to work with `xs={12} sm={6}` on the two columns so it displays on two separate lines if its smaller e.g. image It didn't work properly with just that change as I also had to set the card view height to 0px again. It also likely needs usage of a breakpoint to do this to change the justifyContent in the grid for the pagingation e.g.
> 
> ```
> sx={(theme) => ({
>     [theme.breakpoints.up('sm')]: { justifyContent: 'flex-end' },
> })}
> ```

I think I have improved the behaviour in the latest commit.

joelvdavies (approved):

joshuadkitenge (approved):

PR #472, opened 2024-04-03
Usage statues of moving items #456

joshuadkitenge:
## Description

see #456
## Testing instructions

Add a set up instructions describing how the reviewer should test the code

- [ ] Review code
- [ ] Check Actions build
- [ ] Review changes to test coverage
- [ ] {more steps here}

## Agile board tracking

closes #456

PR #484, opened 2024-04-09
Improve error states #480

### LINKED ISSUE #480: Improve errorstates for system usage statuses
Also just a suggestion, is it possible to store error states perhaps like
errorStates = { item_id: { error: , .catalogue_item_id, ...} }
so they can be looked up with errorState[item_id] and only added to the state if it has an error and deleted if it doesn't. This might be able to avoid a lot of the maps and filters when looking them up here and in the table and should make errorUsageStatusesItemId redundant.

#472

joshuadkitenge:
## Description

see #480

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #480


joelvdavies (commented):
Just one thing I think the errors need to be reset on close. I can move one item, trigger an error, cancel, select a different item in the same system and it will still show an error. Also not sure what was happening with https://github.com/ral-facilities/inventory-management-system/pull/472#discussion_r1553174823.

joelvdavies (approved):

PR #491, opened 2024-04-12
catalogue item property templates #487

### LINKED ISSUE #487: catalogue item property templates
add autocomplete drop down to catalogue item property name textfield in the catalogue catalogue categories dialog

joshuadkitenge:
## Description

- Change the property name to be a freeSolo autocomplete 
- Click an option in the auto complete should updated the fields with the template
- Clear errors on catalogue item field when option is triggered 

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #487

### DIFF
--- a/src/catalogue/category/cataloguePropertiesForm.component.tsx
+++ b/src/catalogue/category/cataloguePropertiesForm.component.tsx
@@ -23,6 +23,7 @@ import {
 } from '../../app.types';
 import { useUnits } from '../../api/units';
 import { generateUniqueId } from '../../utils';
+import { useCatalogueItemPropertyTemplates } from '../../api/catalogue_item_property_templates';
 
 export interface CataloguePropertiesFormProps {
   formFields: CatalogueCategoryFormDataWithPlacementIds[];
@@ -52,6 +53,8 @@ function CataloguePropertiesForm(props: CataloguePropertiesFormProps) {
   } = props;
 
   const { data: units } = useUnits();
+  const { data: catalogueItemPropertyTemplates } =
+    useCatalogueItemPropertyTemplates();
 
   const handleAddField = () => {
     onChangeFormFields([
@@ -295,7 +298,7 @@ function CataloguePropertiesForm(props: CataloguePropertiesFormProps) {
     );
 
     if (fieldIndex === -1) {
-      return; 
+      return;
     }
 
     const updatedFormFields: CatalogueCategoryFormDataWithPlacementIds[] = [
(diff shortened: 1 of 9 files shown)


src/catalogue/category/cataloguePropertiesForm.component.tsx
```diff
@@ -390,25 +393,111 @@ function CataloguePropertiesForm(props: CataloguePropertiesFormProps) {
...
+                  const updatedAllowedValuesListErrors =
+                    allowedValuesListErrors.filter(
+                      (item) => item.cip_placement_id !== field.cip_placement_id
+                    );
+
+                  onChangeAllowedValuesListErrors(
+                    updatedAllowedValuesListErrors
+                  );
+                }
+              }}
```
joelvdavies, line 477: I think it would be good to apply you're suggestion to avoid repeating this logic.

joelvdavies (commented):
Worked well for me. Just a note to reuse function you mentioned earlier.

src/catalogue/category/cataloguePropertiesForm.component.tsx
joshuadkitenge: I don't think what I suggested yesterday is going to work well. The functions in this file currently work well for adding, deleting, updating one field at a time and this onChange updates many fields at time.

joelvdavies: Ok, and its also not sensible to put it into a function to do one update, and then just repeat it for each field for the onChange?

src/catalogue/category/cataloguePropertiesForm.component.tsx
```diff
@@ -390,25 +393,111 @@ function CataloguePropertiesForm(props: CataloguePropertiesFormProps) {
...
+                  (template) => template.name === field.name
+                ) || null
+              }
+              inputValue={field.name}
+              onInputChange={(_event, newValue) => {
+                handleChange(field.cip_placement_id, 'name', newValue);
+              }}
+              onChange={(_event, newValue) => {
+                if (typeof newValue !== 'string' && newValue !== null) {
+                  const updatedFormFields = [...formFields];
```
joelvdavies, line 420: Just a note - I don't believe this prevents updates on the formFields itself as it will only do a shallow copy, not sure if that was the intention though.

joshuadkitenge: this should be a separate issue as it used in multiple places to change it to use `JSON.parse(JSON.stringify(array))`

src/catalogue/category/cataloguePropertiesForm.component.tsx
joshuadkitenge: That's over complicates things. This onChange is simple at add all and clear all errors

src/catalogue/category/cataloguePropertiesForm.component.tsx
joshuadkitenge: #503

src/catalogue/category/cataloguePropertiesForm.component.tsx
joshuadkitenge: to use the current functions in this file you would have to use the handleAdd and then handleChange. This problem because more problematic for the allowed values (loop through a long list doing handleAdd and handleChange)

PR #505, opened 2024-04-16
changed table to use system name and not id #497

### LINKED ISSUE #497: Switch to using System name instead of system id
The current item object only gives back system_id. To display the system name, we would have to fetch for each system id for an item. This could cause performance bottlenecks in the front end. This would be a quicker implement (~ 2 days)

MatteoGuarnaccia5:
## Description

Changed items table to display system name rather than system ID. 
System objects are fetched from backend from set of system IDs - set of IDs is created from items data

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

closes #497


src/api/systems.tsx
```diff
@@ -44,6 +45,15 @@ const fetchSystems = async (parent_id?: string): Promise<System[]> => {
   });
 };
 
+export const useSystemIds = (ids: string[]): UseQueryResult<System>[] => {
```
joelvdavies, line 48: I think this should have a unit test like `useCatalogueItemIds`/`useManufacturerIds`

joelvdavies (changes requested):
Two minor comments. Also could you update the landing pages and detail panels to use system name as well please.

I also noticed that navigating using the system name link from the items table to a systems page causes
image

joshuadkitenge:
Can the system be grouped by the system id and not the name please

joelvdavies (approved):
LGTM

PR #554, opened 2024-05-14
fix duplicate name error for copyTo #553

### LINKED ISSUE #553: fix duplicate names (all dialogs)
Check if a you can change the name (duplicate value) to the same name with capital letter 
- All dialog work correct with mock data requires backend change to work 
    - works 

Check move to and copy to

- Fix the systems copy to and the systems copy to
    - they need to use code instead of name

joshuadkitenge:
## Description

fix the copy to and move to for the directory dialogs. Use the code for the duplicate check instead of name.  

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #553


MatteoGuarnaccia5 (dismissed):
LGTM

joelvdavies (changes requested):
I only just realised we seem to have missed out tests on duplicates. I think it would be good if they could be added here.

joelvdavies:
Is it possible to check the moves as well with the same tests you added? We are missing those too.

joshuadkitenge:
> Is it possible to check the moves as well with the same tests you added? We are missing those too.

The move doesn't have the duplicate same logic . The failed response ( for duplicate) is tested in the hook itself catalogueCategories.test.tsx

MatteoGuarnaccia5 (approved):
LGTM

joelvdavies (approved):

Issue #599, opened 2024-05-30
row totals do not display correct info if no network connection


MatteoGuarnaccia5:
![image](https://github.com/ral-facilities/inventory-management-system/assets/144466888/a916628a-dede-4885-9cfb-356892d4f0ae)

Should really be `Total: 0 Systems` - same issue occurs on every table

PR #603, opened 2024-05-31
Preserved table filters fixes #597

### LINKED ISSUE #597: preserving filter state of datetimes does not work
If you filter any of the tables (example below) by a datetime field, then send that URL link. It does not open the correct table state.
The filtered field has an error, after clearing the `Clear filters` button no longer works and stays active even after clearing all filters.

Animation

### LINKED ISSUE #610: System Items Table Loses Catalogue Item Name Grouping After Refresh
https://github.com/ral-facilities/inventory-management-system/assets/83226114/2180b495-1d2a-4fe5-99ee-f0c6df3a8271

joelvdavies:
## Description

See #597

- Fixes bug where min/max filters wouldn't be removed from the URL when they were cleared #597
- Fixes bug where date filters could not be loaded from the URL #599
- Fixes bug where drag and drop column grouping wouldn't  update the URL so refreshing would reset #610 

Added an additional unit and e2e test for catalogue items that would have failed without these fixes.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Ensure table filters work correctly for different types e.g. numbers, dates, strings, booleans

## Agile board tracking

Closes #597, Closes #610


src/common/preservedTableState.component.test.tsx
```diff
@@ -150,7 +177,7 @@ describe('Preserved table state functions', () => {
       // };
       // console.log(
```
MatteoGuarnaccia5, line 178: is the `console.log` needed?

MatteoGuarnaccia5 (commented):
LGTM - just one thing

MatteoGuarnaccia5 (approved):
Works well for me

joshuadkitenge (changes requested):
the fix for #610   works well.

I found this bug for date range filters 

https://github.com/ral-facilities/inventory-management-system/assets/83226114/e77ef6d9-a668-48f9-ab96-b46cf21f95e2




I have also noticed that some weird behavior with the range filters in MRT. this a MRT problem.  I've create another issue for this to add custom error handling.#634
image

PR #525, opened 2024-04-29
changed selects to be autocompletes in cat Items and items dialogs #495

MatteoGuarnaccia5:
## Description

Changed the select inputs in catalogue items and items dialog to be autocompletes

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Review changes to test coverage

## Agile board tracking

closes #495

PR #616, opened 2024-06-05
added ids to all text fields #613

### LINKED ISSUE #613: Add ids to text fields and autocompletes
Found this on chrome and discussed with Joshua. Effectively chrome by default will save any `<input>` elements when a form is submitted and when editing an `<input>` field with the same `id` it will display those same options to the user. Firefox and Edge both seem to ignore these, I assume because it only does it for password or payment information. 

The issue here is that the id's in all text fields and autocompletes are generated automatically by MUI and frequently clash so that fields start suggesting things that were previously entered into completely different fields e.g.
image, this can also change on page refresh.

While we can technically disable this using `autocomplete="off"` in all text fields (which appears to have not worked for the past 8 years on Chrome according to https://issues.chromium.org/issues/41239842 and https://mui.com/material-ui/react-autocomplete/#autocomplete-autofill, but worked when we tested it today), this is not overridable for autocompletes in MRT which currently uses `autocomplete="new-password"` which doesn't work for Chrome and is just a hint rather than a rule.

So overall it seems we must assign ids to all TextField's that are descriptive enough to avoid these conflicts. I believe autocompletes themselves outside of MRT may be alright, but can probably do with ids anyway to match the rest.

MatteoGuarnaccia5:
## Description

see #613

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

closes #613


MatteoGuarnaccia5:
wait on #525 to be merged so that IDs can be added to autocompletes

joelvdavies (commented):
A few minor suggestions. Looks like it has resolved the original issue 🥳

src/catalogue/category/__snapshots__/cataloguePropertyForm.component.test.tsx.snap
```diff
@@ -2267,8 +2267,8 @@ exports[`Catalogue Property Form > renders correctly when disabled 1`] = `
         <label
           class="MuiFormLabel-root MuiInputLabel-root MuiInputLabel-formControl MuiInputLabel-animated MuiInputLabel-sizeMedium MuiInputLabel-outlined MuiFormLabel-colorPrimary Mui-disabled MuiInputLabel-root MuiInputLabel-formControl MuiInputLabel-animated MuiInputLabel-sizeMedium MuiInputLabel-outlined css-14s5rfu-MuiFormLabel-root-MuiInputLabel-root"
           data-shrink="false"
-          for=":ri:"
-          id=":ri:-label"
+          for="undefined-propety-unit-input"
+          id="undefined-propety-unit-input-label"
```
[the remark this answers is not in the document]
MatteoGuarnaccia5: I had to try to find a way to make each id unique, when there is multiple properties in the form. Otherwise for some reason the labels all duplicate onto only the first textfield. I didn't realise `cip_placement_id` could be undefined, I'll find a fix for it

src/catalogue/category/__snapshots__/cataloguePropertyForm.component.test.tsx.snap
```diff
@@ -272,8 +272,8 @@ exports[`Catalogue Property Form > renders correctly 1`] = `
         <label
           class="MuiFormLabel-root MuiInputLabel-root MuiInputLabel-formControl MuiInputLabel-animated MuiInputLabel-sizeMedium MuiInputLabel-outlined MuiFormLabel-colorPrimary MuiInputLabel-root MuiInputLabel-formControl MuiInputLabel-animated MuiInputLabel-sizeMedium MuiInputLabel-outlined css-14s5rfu-MuiFormLabel-root-MuiInputLabel-root"
           data-shrink="false"
-          for=":r4:"
-          id=":r4:-label"
+          for="property-unit-input"
```
joelvdavies, line 275: I think the unit tests may need `props.cip_placement_id` so that it adds the index as it should be there.

src/catalogue/category/cataloguePropertyForm.component.tsx
```diff
@@ -329,6 +333,7 @@ function CataloguePropertyForm(props: CataloguePropertyFormProps) {
           />
         ))}
       <Autocomplete
+        id={`property-unit-input${cip_placement_id ?? ''}`}
```
joelvdavies, line 336:
```suggestion
        id={`property-unit-input-${cip_placement_id ?? ''}`}
```

src/systems/systemItemsTable.component.tsx
```diff
@@ -445,7 +445,7 @@ export function SystemItemsTable(props: SystemItemsTableProps) {
                 return (
                   <FormControl size="small" fullWidth>
                     <Autocomplete
-                      id={`usage-statuses-${row.original.catalogueItem?.name}`}
+                      id={`usage-statuses-${row.original.item?.serial_number}`}
```
joshuadkitenge, line 448:
```suggestion
                      id={`usage-statuses-${row.original.item?.serial_number ?? 'No ...'}`}
```

joelvdavies:
Testing this I found it was inconsistent with property ids as they are generated while the page is open e.g. adding two categories wont remember the first. Discussing with Joshua we think its best just to disable it for these by generating unique random ids for them rather than the ones generated with a counter. I think you can use the `uuid` package to do this and use uuidv4 (cypress already includes this, but we will need to add it to the non dev dependencies with `yarn add uuid @types/uuid`).

MatteoGuarnaccia5:
> Testing this I found it was inconsistent with property ids as they are generated while the page is open e.g. adding two categories wont remember the first. Discussing with Joshua we think its best just to disable it for these by generating unique random ids for them rather than the ones generated with a counter. I think you can use the `uuid` package to do this and use uuidv4 (cypress already includes this, but we will need to add it to the non dev dependencies with `yarn add uuid @types/uuid`).

Would just be in the propertyForm ? Or everywhere in which there are properties, like in the catalogue items / items dialog.

joelvdavies:
> > Testing this I found it was inconsistent with property ids as they are generated while the page is open e.g. adding two categories wont remember the first. Discussing with Joshua we think its best just to disable it for these by generating unique random ids for them rather than the ones generated with a counter. I think you can use the `uuid` package to do this and use uuidv4 (cypress already includes this, but we will need to add it to the non dev dependencies with `yarn add uuid @types/uuid`).
> 
> Would just be in the propertyForm ? Or everywhere in which there are properties, like in the catalogue items / items dialog.

Items form looks fine to me as it uses the ids of the properties. The catalogue items form currently use the name, but should use id as well if you can change that. The migration add property form currently renders with names like `property-name-input-` and as they have the same issue with the allowed values list, I think they should as well but that is in the propertyForm anyway. So I think it is just the propertyForm that needs the uuid's.

src/catalogue/category/cataloguePropertiesForm.component.test.tsx
```diff
@@ -49,6 +50,9 @@ describe('Catalogue Properties Form', () => {
   });
 
   it('renders correctly', async () => {
+    vi.mock('uuid', () => {
+      return { v4: vi.fn(() => '00000000-0000-0000-0000-000000000000') };
```
joelvdavies, line 54: Could you use something like this in `testUtils` so that they don't all use the same ids.
```
/* This counter and the function below it are used for forcing UUID's to be deterministic */
let globalUUIDCounter = 0;

const generateUUIDBytes = (): Uint8Array => {
  const byteArray = new Uint8Array(16);
  for (let i = 0; i < byteArray.length; i++) {
    byteArray[i] = (globalUUIDCounter + i) % 256; // Ensuring the value stays within byte range
  }
  globalUUIDCounter++;
  return byteArray;
};

const mockUUIDv4 = () => {
    // Mock uuid generation in all tests to ensure they are deterministic rather than random
  // for snapshots
  globalUUIDCounter = 0;
  vi.mock('uuid', async () => {
    const uuid = await vi.importActual('uuid');
    return {
      ...uuid,
      v4: vi.fn(() =>
        // @ts-expect-error Unknown as don't want to import here
        uuid.v4({
          random: generateUUIDBytes(),
        })
      ),
    };
  });
}
```

src/catalogue/category/cataloguePropertiesForm.component.test.tsx
```diff
@@ -29,6 +29,7 @@ describe('Catalogue Properties Form', () => {
   };
 
   beforeEach(() => {
+    vi.resetAllMocks();
```
joelvdavies, line 32: This is normally put into the `afterEach`.

MatteoGuarnaccia5: The snapshots don't seem to work when I put in the afterEach, thats why I put it in the beforeEach

MatteoGuarnaccia5: When putting it in the `afterEach` the snapshot doesn't expect ids anymore, they are all like `rif0` instead of a byte string; which to me is wrong

MatteoGuarnaccia5: image

This is with `vi.resetAllMocks()` in the afterEach. To me that does not seem right

joelvdavies: That is saying the snapshot contains ids like `:r1f:` but it is getting uuidv4's instead, which is what we want. Letting it updates makes these changes:
image

MatteoGuarnaccia5: Yeah I just realised that - my bad I got confused with what was actually correct

joelvdavies:
Could the case match on the selects, e.g. right now it shows they are different e.g.
image

Also could the clear button be removed on the fields that are mandatory - it doesn't do anything after a value has been selected as they can't be cleared.

MatteoGuarnaccia5:
> Could the case match on the selects, e.g. right now it they are different e.g. image
> 
> Also could the clear button be removed on the fields that are mandatory - it doesn't do anything after a value has been selected as they can't be cleared.

I was going to create an issue to change the autocompletes to ensure the labels and value work better - I will probably do it in #643, and can get rid of the clear option if mandatory

joshuadkitenge (approved):
LGTM

joelvdavies (approved):

PR #675, opened 2024-06-17
Rename manufacturer api functions #674

joshuadkitenge:
## Description

Rename api function to includes the http method 

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

#674


src/api/manufacturers.tsx
```diff
@@ -106,15 +106,15 @@ const editManufacturer = async (
...
 
-export const useEditManufacturer = (): UseMutationResult<
+export const usePatchManufacturer = (): UseMutationResult<
   Manufacturer,
   AxiosError,
   ManufacturerPatch
 > => {
   const queryClient = useQueryClient();
   return useMutation({
     mutationFn: (manufacturer: ManufacturerPatch) =>
```
joelvdavies, line 116: Now that the types match very closely, I would kind of prefer this to take an id and the patch as id isn't really part of the patch schema. But it could also be more of an audit thing and would be best addressed after the form changes.

joelvdavies (approved):
Minor comment, but would only effect later. LGTM

src/api/manufacturers.tsx
```diff
@@ -70,34 +70,34 @@ export const useDeleteManufacturer = (): UseMutationResult<
...
+      queryFn: () => getManufacturer(id),
     })),
   });
 };
 
-const editManufacturer = async (
+const patchManufacturer = async (
   manufacturer: ManufacturerPatch
 ): Promise<Manufacturer> => {
   const { id, ...updatedManufacturer } = manufacturer;
```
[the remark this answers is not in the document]
joshuadkitenge: and the relevant change to usePatchManufacturer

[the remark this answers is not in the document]
joelvdavies: Yeah exactly, perhaps id as the first argument. Then calling patch requires the id of the entity being patched as well as the patch data instead while keeping the patch schema as just the schema of the patch request body itself.

src/api/manufacturers.tsx
joshuadkitenge: I get what you mean but the way it is currently done now doesn't decrease readability. The main type that we need to make sure is exactly as the backend is the output type (Manufacturer,System) all of the other (post and patch) ones should contain the information we need the react query hook to send the information

joelvdavies: Yeah that's fine, its just me being picky with the definitions. Saying its the api schema implies to me it should be what the api takes, not the front end function that sends it otherwise it would be in the app.types as its an app type not an api type. But I am fine with it here now.

joelvdavies (approved):

PR #688, opened 2024-06-18
 fix type interfaces units #687

### LINKED ISSUE #687: Fix usage status type interfaces

joshuadkitenge:
## Description

- Rename so it has the http method in the name
- Move the types use in react query functions to api.types.tsx

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #687


src/api/api.types.tsx
```diff
@@ -56,3 +56,13 @@ export interface Unit extends UnitPost, CreatedModifiedMixin {
   id: string;
   code: string;
 }
+
+// Usage Statuses
```
joshuadkitenge, line 60:
```suggestion
// ------------------------------------ Usage Status ------------------------------------------------
```

src/api/api.types.tsx
```diff
@@ -56,3 +56,13 @@ export interface Unit extends UnitPost, CreatedModifiedMixin {
   id: string;
   code: string;
 }
+
+// ------------------------------------ Usage Status ------------------------------------------------
```
joelvdavies, line 60: I think either this and handlers should use all caps or otherwise all of them should be like this.

joshuadkitenge: I noticed this in the systems one and I have changed it there (#697) also it is inconsistent in the handlers so that need to be addressed in the audit

joelvdavies: As its such a small I would just do it there now if possible. Its possible we could miss these things if we don't make a note.

joelvdavies (approved):

PR #699, opened 2024-06-19
refactored external links test #689

### LINKED ISSUE #689: Refactor external link tests on cypress
In our current end-to-end tests, we navigate to a page to verify that a link works. However, this approach causes issues if the external page is down. Instead, we should simply check that the element has the correct href and target attributes using Cypress tests.

```
    // Find the link element
    cy.findAllByText('http://example-drawing-link.com')
      .first()
      .should('have.attr', 'href')
      .should('include', 'http://example-drawing-link.com'); // Check href attribute value

    cy.findAllByText('http://example-drawing-link.com')
      .first()
      .should('have.attr', 'target')
      .should('include', '_blank'); // Check target attribute value
```

MatteoGuarnaccia5:
## Description

refactored tests to check url attributes rather than link to url

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

closes #689

### DIFF
--- a/cypress/e2e/with_mock_data/homePage.cy.ts
+++ b/cypress/e2e/with_mock_data/homePage.cy.ts
@@ -16,16 +16,12 @@ describe('IMS HomePage', () => {
     cy.url().should('include', '/manufacturers');
     cy.go('back');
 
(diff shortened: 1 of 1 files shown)


joshuadkitenge (approved):
LGTM

PR #718, opened 2024-06-27
Fix type interfaces for catalogue categories #710

joshuadkitenge:
## Description

Create new type for catalogue categories which match better with the backend schemas 

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

#710


src/api/api.types.tsx
```diff
@@ -101,3 +101,64 @@ export interface System extends CreatedModifiedMixin {
...
+// ------------------------------------ CATALOGUE CATEGORIES ------------------------------------
+
+export interface AllowedValuesList {
+  type: 'list';
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  values: any[];
+}
+export type AllowedValues = AllowedValuesList;
+
+export enum CatalogueCategoryPropertyTypes {
```
joelvdavies, line 114:
```suggestion
export enum CatalogueCategoryPropertyType {
```
Types are usually singular and this matches the backend.

src/api/api.types.tsx
```diff
@@ -101,3 +101,64 @@ export interface System extends CreatedModifiedMixin {
...
+}
+
+export interface CatalogueCategoryPropertyPost
+  extends CatalogueCategoryPostProperty {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  default_value: any;
+}
+
+export interface CatalogueCategoryPropertyPatch {
+  name?: string;
```
joelvdavies, line 135:
```suggestion
  name?: string | null;
```
Then again I assume this is because null is used elsewhere to indicate setting it to null, whereas null here would be the same as undefined?

src/api/api.types.tsx
```diff
@@ -101,3 +101,64 @@ export interface System extends CreatedModifiedMixin {
...
+  id: string;
+  unit_id: string | null;
+  unit: string | null;
+  allowed_values: AllowedValues | null;
+}
+
+export interface CatalogueCategoryPost {
+  name: string;
+  parent_id?: string | null;
+  is_leaf: boolean;
```
joelvdavies, line 150:
```suggestion
   is_leaf: boolean;
  parent_id?: string | null;
```
just to match.

src/api/api.types.tsx
```diff
@@ -101,3 +101,64 @@ export interface System extends CreatedModifiedMixin {
...
+  properties?: CatalogueCategoryPostProperty[];
```
joelvdavies, line 151:
```suggestion
  properties?: CatalogueCategoryPostProperty[] | null;
```

src/api/api.types.tsx
```diff
@@ -101,3 +101,64 @@ export interface System extends CreatedModifiedMixin {
...
+export interface CatalogueCategoryPatch
+  extends Partial<CatalogueCategoryPost> {}
+
+export interface CatalogueCategory
+  extends Omit<CatalogueCategoryPost, 'parent_id' | 'properties'>,
+    CreatedModifiedMixin {
+  id: string;
+  code: string;
+  parent_id: string | null;
+  properties: CatalogueCategoryProperty[];
```
joelvdavies, line 163:
```suggestion
  properties: CatalogueCategoryProperty[] | null;
```

src/api/api.types.tsx
joshuadkitenge: This is not consistent with what was done in the previous prs

src/api/api.types.tsx
joshuadkitenge: This change is not consistence with the backend. if i send back nothing it's give back [ ] (empty list) for the properties

src/api/api.types.tsx
joshuadkitenge: yeah null is the same as undefined in this case

src/api/api.types.tsx
joelvdavies: Yeah ok, if anything the backend schema is wrong, the empty list is defined in the database models.

joelvdavies (approved):

joelvdavies (approved):

PR #892, opened 2024-08-16
Add Recently Added Section to Manufacturer Autocomplete #717

### LINKED ISSUE #717: Add a "Recently Added" Section to Manufacturer Autocomplete
### Description:

Enhance the manufacturer autocomplete feature by adding a "Recently Added" section. The criteria for "Recently Added" should be items added within the last 5 or 10 minutes. These recently added items should be displayed in order of the most recent ones first.

asuresh-code:
## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

connect to #717 
closes #717


cypress/e2e/with_mock_data/catalogueItems.cy.ts
```diff
@@ -404,6 +405,17 @@ describe('Catalogue Items', () => {
...
+    cy.clock(new Date("2024-01-09T12:00:00.000+00:00"), ['Date']);
+    
+    cy.findByText('Cameras 1').click();
+    cy.findByRole('button', { name: 'Edit' }).click();
+    cy.findByLabelText('Manufacturer *').click();
+    cy.contains('A-Z').should('be.visible');
+    cy.contains('Recently Added').should('be.visible');
+    cy.findAllByText('Manufacturer B').should('have.length', 2);
+  })
+
```
joshuadkitenge, line 418: Can you restore the after the test has finished please. https://docs.cypress.io/api/commands/clock#Restore-clock

src/catalogue/items/catalogueItemsDialog.component.tsx
```diff
@@ -605,6 +605,37 @@ function CatalogueItemsDialog(props: CatalogueItemsDialogProps) {
...
+      };
+    }) : [];
+
+    //duplicating the recent manufacturers, so that they appear twice.
+    const date_now = (new Date()).getTime()
+    const recentManufacturers = classifiedManufacturers
+    .filter((option) => {
+      const created_date = (new Date(option.created_time)).getTime()
+      //dates are in ms; numerical expression is equivalent to 10 minutes
+      const isRecent = date_now - 1000*60*10 <= created_date
```
joshuadkitenge, line 623: Can you change the recently added time to a variable please

src/catalogue/items/catalogueItemsDialog.component.test.tsx
```diff
@@ -148,6 +148,7 @@ describe('Catalogue Items Dialog', () => {
 
   afterEach(() => {
     vi.clearAllMocks();
+    vi.useRealTimers();
```
joshuadkitenge, line 151:
```suggestion
```
You already have this at the end of the test ('displays recently added section')

src/catalogue/items/catalogueItemsDialog.component.test.tsx
```diff
@@ -621,6 +622,27 @@ describe('Catalogue Items Dialog', () => {
     expect(onClose).not.toHaveBeenCalled();
   }, 10000);
 
+  it('displays recently added section', async () => {
+    vi.setSystemTime(new Date("2024-01-09T12:00:00.000+00:00"));
```
joshuadkitenge, line 626:
```suggestion
    vi.useFakeTimers()
    vi.setSystemTime(new Date("2024-01-09T12:00:00.000+00:00"));
```

From this documentation (https://vitest.dev/guide/mocking.html#example) is looks like you should be using `vi.useFakeTimers()` before running the test.

joshuadkitenge (changes requested):
Works well, just a couple of minor comments

joelvdavies (changes requested):
Haven't looked at the code yet, but noticed that clicking the add manufacturer button again after using it to create a new manufacturer doesn't clear the previous content, this doesn't seem to have been the case before this PR.

src/catalogue/items/catalogueItemsDialog.component.test.tsx
joelvdavies: If its only in this test I think this is fine rather than mocking the time for each test when we don't need to.

src/catalogue/items/catalogueItemsDialog.component.test.tsx
asuresh-code: Is it fine to remove it from the end of the test instead? In previous runs when a test failed, it didn't reach the vi.useRealTimers() cmd, and caused all subsequent tests to fail.

src/catalogue/items/catalogueItemsDialog.component.test.tsx
asuresh-code: I tried using useFakeTimers before, and was having difficulties advancing time around an await function, however I was able to fix the issue now.

Also, I have added useRealTimers in the afterEach section of the test file.

cypress/e2e/with_mock_data/catalogueItems.cy.ts
asuresh-code: Is it fine to restore it in the afterEach section, for the same reason mentioned here? https://github.com/ral-facilities/inventory-management-system/pull/892#discussion_r1742284330

joelvdavies: As its only needed for this test alone, I would only restore in this test and mock it in this test instead of using the beforeEach, the same goes for the unit tests.

asuresh-code:
> Haven't looked at the code yet, but noticed that clicking the add manufacturer button again after using it to create a new manufacturer doesn't clear the previous content, this doesn't seem to have been the case before this PR.

I've been able to recreate the bug on the develop branch, so I don't think it's an issue with this PR.

joelvdavies:
> > Haven't looked at the code yet, but noticed that clicking the add manufacturer button again after using it to create a new manufacturer doesn't clear the previous content, this doesn't seem to have been the case before this PR.
> 
> I've been able to recreate the bug on the develop branch, so I don't think it's an issue with this PR.

Yep you're right, Joshua fixed it in https://github.com/ral-facilities/inventory-management-system/pull/897.

src/catalogue/items/catalogueItemsDialog.component.test.tsx
joelvdavies: This will repeatedly runs it after every test when its only needed for the one test - I just tried moving`vi.useRealTimers();` to the end of the `'displays recently added section'` test and the file worked fine for me so I would do that as its just that test that needs to mock/unmock the timers.

src/catalogue/items/catalogueItemsDialog.component.test.tsx
```diff
@@ -621,6 +622,27 @@ describe('Catalogue Items Dialog', () => {
...
+    createView();
+    const manufacturerPopup = screen.getAllByRole('combobox')[0];
+    
+    user.type(manufacturerPopup, 'Man')
+    vi.advanceTimersByTimeAsync(2000);
+    const options = await screen.findAllByRole('option');
+    expect(options).toHaveLength(5)
+    expect(screen.getAllByText("Manufacturer B")).toHaveLength(2);
+    expect(screen.getByText("A-Z")).toBeInTheDocument()
+    expect(screen.getByText("Recently Added")).toBeInTheDocument()
```
joelvdavies, line 643: Some of these have semi colons some don't, could be a vscode setting. If you do ctrl+shift+p and type Format with... you can select a default formatter of prettier (you may need the extension). We just have it set to format on save - again if you do ctrl+shift+p and search for settings, then in the settings search for format on save and check it then it should do it automatically.

image
image

src/catalogue/items/catalogueItemsDialog.component.test.tsx
asuresh-code: My main concern is if there's an error before it reaches  `vi.useRealTimers();` For instance if you comment out `vi.advanceTimersByTimeAsync(2000);` this happens.
image

I tried using a try, finally block in the test to run useRealTimers(), however the test ignores it when it times out. Would it be fine to put this test in its own describe block, and use `afterEach `and `beforeEach` for just this test.

joelvdavies: Ah yeah good spot, I didn't think of that, yeah I would personally just do that so it has a set of tests with a mocked clock. Same for the e2e tests so that the others can keep the real system time.

src/catalogue/items/catalogueItemsDialog.component.test.tsx
asuresh-code: Ah, thanks for the advice. I had just been doing it manually, and missing it in a few places. Got it enabled now.

src/catalogue/items/catalogueItemsDialog.component.tsx
```diff
@@ -605,6 +605,37 @@ function CatalogueItemsDialog(props: CatalogueItemsDialogProps) {
...
+    .filter((option) => {
+      const created_date = (new Date(option.created_time)).getTime()
+      //dates are in ms; numerical expression is equivalent to 10 minutes
+      let isRecent = date_now - 1000*60*10 <= created_date
+      return isRecent;
+    })
+    .map((option) => {
+      return {
+        ...option,
+        isrecent: "Recently Added",
```
joshuadkitenge, line 629:
```suggestion
        isRecent: "Recently Added",
```

src/catalogue/items/catalogueItemsDialog.component.tsx
```diff
@@ -605,6 +605,37 @@ function CatalogueItemsDialog(props: CatalogueItemsDialogProps) {
...
+      };
+    }) : [];
+
+    //duplicating the recent manufacturers, so that they appear twice.
+    const date_now = (new Date()).getTime()
+    const recentManufacturers = classifiedManufacturers
+    .filter((option) => {
+      const created_date = (new Date(option.created_time)).getTime()
+      //dates are in ms; numerical expression is equivalent to 10 minutes
+      let isRecent = date_now - 1000*60*10 <= created_date
```
joshuadkitenge, line 623: There are some semi colons missing in this function. 
Can you change this number into a variable please `1000*60*10`


https://github.com/ral-facilities/inventory-management-system/pull/892#discussion_r1740607006
https://github.com/ral-facilities/inventory-management-system/pull/892#discussion_r1743507149

src/catalogue/items/catalogueItemsDialog.component.tsx
```diff
@@ -605,6 +605,37 @@ function CatalogueItemsDialog(props: CatalogueItemsDialogProps) {
     [errorMessages, propertyErrors]
   );
 
+
+  const options = ():Array<Manufacturer & {isrecent: string}> => {
+    const classifiedManufacturers = manufacturerList ? (manufacturerList).map((option) => {
+      return {
+        ...option,
+        isrecent: "A-Z"
```
joshuadkitenge, line 613:
```suggestion
        isRecent: "A-Z"
```

src/catalogue/items/catalogueItemsDialog.component.tsx
```diff
@@ -794,7 +825,8 @@ function CatalogueItemsDialog(props: CatalogueItemsDialogProps) {
                       newManufacturer?.id ?? null
                     );
                   }}
-                  options={sortDataList(manufacturerList ?? [], 'name')}
+                  options={sortDataList(options(), "isrecent").reverse() ?? []}
```
joshuadkitenge, line 828:
```suggestion
                  options={sortDataList(options(), "isRecent").reverse() ?? []}
```

joshuadkitenge (changes requested):

src/catalogue/items/catalogueItemsDialog.component.tsx
asuresh-code: Cheers for pointing it out, added them here and in other files now.
Ah right, sorry misunderstood you in the earlier comment.

src/catalogue/items/catalogueItemsDialog.component.tsx
```diff
@@ -605,6 +605,39 @@ function CatalogueItemsDialog(props: CatalogueItemsDialogProps) {
...
+          };
+        })
+      : [];
+
+    //duplicating the recent manufacturers, so that they appear twice.
+    const date_now = new Date().getTime();
+    const recentManufacturers = classifiedManufacturers
+      .filter((option) => {
+        const created_date = new Date(option.created_time).getTime();
+        const TEN_MINUTES = 1000 * 60 * 10;
```
joshuadkitenge, line 623: Can you move `const TEN_MINUTES = 1000 * 60 * 10;` to the top of the file below the imports please

joelvdavies: I think it would also be nice if instead of being `TEN_MINUTES` it was named something more specific to what its for e.g. `RECENT_MANUFACTURER_MAX_AGE_MINUTES`?

joelvdavies (approved):
LGTM, Thanks for this, it worked well for me

joshuadkitenge (dismissed):

joelvdavies (approved):

joshuadkitenge (approved):

PR #922, opened 2024-09-03
Enhance Usage Statuses Table Filtering

asuresh-code:
## Description

Value:
- Filter variant: Text
_Data is string which leaves 5 options: select, checkbox, autocomplete, multiselect and text. Checkbox only allows simple yes/no filtering, so not suitable. Select and Multi-select don't allow users to type, which is restrictive. All records have unique values, therefore autocomplete is less useful, though still usable. Ended up picking text as its the most flexible variant._
- Filter mode: [Fuzzy, Contains, Starts With, Ends With, Not Equals, Between Inclusive]
_The first 4 are the most useful filter functions for filtering string data. Not equals could be a useful filter for filtering out a value. Between inclusive offers all of the functionality of between,<=,>=, etc, which could be useful to get ranges of text. Every other filter could either be recreated from the chosen 6, or is superfluous._

Last Modified/Created:
- Filter variant = Datetime-Range
_Datetime is chosen since the data has that level of specificity. Range is chosen since it allows more functionality, and filtering for an exact datetime is not needed._
- Filter mode = ['between inclusive', 'between']:
_The only 2 range filters. Between doesn't add much value since it is only omitting values by 1 second (i.e a lower bound of 8/09/23 13:23 will include 8/09/23 13:24), but is left in since (1) the user may still find use for it, (2) There are only 2 filter functions, so it would not be overwhelming the user by including it, and (3) they are MRT functions, so do not need to be actively maintained._

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

connect to #979


src/common/preservedTableState.component.tsx
```diff
@@ -367,6 +374,25 @@ export const usePreservedTableState = (props?: UsePreservedTableStateProps) => {
...
+        const isDefaultState =
+          JSON.stringify(initialValue) === JSON.stringify(newValue);
+        return {
+          ...prevState,
+          cFn: isDefaultState ? undefined : newValue,
+        };
+      });
+    },
+    [defaultState.cFn, updateSearchParams]
+  );
```
joelvdavies, line 394: Could the unit tests be updated to cover this as well? See `onColumnFiltersChange updates the state and url correctly` and similar.

src/utils.test.tsx
```diff
@@ -292,6 +296,38 @@ describe('Utility functions', () => {
   });
 });
 
+describe('Custom Filter Functions', () => {
+  it('correctly removes seconds from date', () => {
```
joelvdavies, line 300:
```suggestion
  it('removeSecondsFromDate correctly removes seconds from date', () => {
```
Just so its clear what it tests. The others have the function name in the describe instead.

src/utils.test.tsx
```diff
@@ -292,6 +296,38 @@ describe('Utility functions', () => {
...
+    const inputDate: string = '2024-01-02T13:10:10.000+00:00';
+    const expectedResult: Date = new Date('2024-01-02T13:10:00.000+00:00');
+    const actualResult: Date = removeSecondsFromDate(inputDate);
+    expect(actualResult).toEqual(expectedResult);
+  });
+});
+
+it('correctly creates filterFns initial state', () => {
```
joelvdavies, line 308:
```suggestion
it('getInitialColumnFilterFnState correctly creates filterFns initial state', () => {
```

joelvdavies (commented):
Just comments on tests, I think some are missing looking at codecov but the URL states worked well for me 🥳

joelvdavies:
> * Filter mode: [Fuzzy, Contains, Starts With, Ends With, Not Equals, Between Inclusive]
>   _The first 4 are the most useful filter functions for filtering string data. Not equals could be a useful filter for filtering out a value. Between inclusive offers all of the functionality of between,<=,>=, etc, which could be useful to get ranges of text. Every other filter could either be recreated from the chosen 6, or is superfluous._

Code LGTM, just one query on the chosen filter functions. I am not sure what between inclusive means for a text field here? Is it doing a comparison based on the ASCII codes?

asuresh-code:
> > * Filter mode: [Fuzzy, Contains, Starts With, Ends With, Not Equals, Between Inclusive]
> >   _The first 4 are the most useful filter functions for filtering string data. Not equals could be a useful filter for filtering out a value. Between inclusive offers all of the functionality of between,<=,>=, etc, which could be useful to get ranges of text. Every other filter could either be recreated from the chosen 6, or is superfluous._
> 
> Code LGTM, just one query on the chosen filter functions. I am not sure what between inclusive means for a text field here? Is it doing a comparison based on the ASCII codes?

It was not clear to me from looking at the implementation of it [found here](https://github.com/KevinVandy/material-react-table/blob/v3/packages/material-react-table/src/fns/filterFns.ts), whether it does it based on ASCII codes. However, that does seem to be the case after I spent some time testing the filter.

It is not case-sensitive, so a lower bound of Z (90) would still exclude lowercase letters like p (112). In the implementation they do convert all string to lowercase before comparison. However, the rest of the characters behave as expected, so numbers (47-58) would include all letters as a lower bound, but exclude all as an upper bound.

src/admin/units/units.component.tsx
```diff
@@ -48,13 +49,15 @@ function Units() {
         header: 'Value',
         accessorFn: (row) => row.value,
         id: 'value',
+        filterFn: 'fuzzy',
```
joshuadkitenge, line 52:
```suggestion
```

src/admin/usageStatuses/usageStatuses.component.tsx
```diff
@@ -49,24 +51,35 @@ function UsageStatuses() {
...
         id: 'value',
+        filterVariant: 'text',
+        filterFn: 'fuzzy',
+        columnFilterModeOptions: [
+          'fuzzy',
+          'contains',
+          'startsWith',
+          'endsWith',
+          'notEquals',
+          'betweenInclusive',
```
joshuadkitenge, line 62:
```suggestion
```
Doesn't make sense for text value

src/admin/usageStatuses/usageStatuses.component.tsx
```diff
@@ -49,24 +51,35 @@ function UsageStatuses() {
...
         accessorFn: (row) => row.value,
         id: 'value',
+        filterVariant: 'text',
+        filterFn: 'fuzzy',
+        columnFilterModeOptions: [
+          'fuzzy',
+          'contains',
+          'startsWith',
+          'endsWith',
+          'notEquals',
```
joshuadkitenge, line 61:
```suggestion
          'notEquals',
          'equals'
```
If you have "noEquals" it should have "equals" as well

src/admin/usageStatuses/usageStatuses.component.tsx
```diff
@@ -49,24 +51,35 @@ function UsageStatuses() {
...
+        ],
         Cell: ({ row }) => row.original.value,
       },
       {
         header: 'Last modified',
-        accessorFn: (row) => new Date(row.modified_time),
+        accessorFn: (row) => removeSecondsFromDate(row.modified_time),
         id: 'modified_time',
         filterVariant: 'datetime-range',
+        filterFn: 'between',
```
joshuadkitenge, line 71:
```suggestion
        filterFn: 'betweenInclusive',
```

src/admin/units/units.component.tsx
```diff
@@ -48,13 +49,15 @@ function Units() {
...
         Cell: ({ row }) => row.original.value,
       },
       {
         header: 'Last modified',
-        accessorFn: (row) => new Date(row.modified_time),
+        accessorFn: (row) => removeSecondsFromDate(row.modified_time),
```
joshuadkitenge, line 57:
```suggestion
        accessorFn: (row) => removeSecondsFromDate(row.modified_time),
```

why is the removed seconds from date needed ??

src/utils.tsx
```diff
@@ -395,6 +398,28 @@ export const displayTableRowCountText = <TData extends MRT_RowData>(
   return <Typography sx={{ ...sx }}>{tableRowCountText}</Typography>;
 };
 
+export const getInitialColumnFilterFnState = (
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  columns: MRT_ColumnDef<any>[]
```
joshuadkitenge, line 403: You should be able to use an template here instead of any

joshuadkitenge (changes requested):
Works well in the url, just a couple of comments

src/admin/units/units.component.tsx
asuresh-code: Previously, dates would be stored to the second, but would only be displayed and filtered to the minute. This creates some unexpected behaviour for users. For instance a value of `01/01/24 13:23 05 seconds` would be excluded with an upper bound of '01/01/24 13:23', since the extra seconds put it out of range. Therefore, I created and added the function to all of the date values in the columns.

joshuadkitenge: That's not unexpected behavior when using a date range, The user should just extend the range to  `01/01/24 13:24` to include all possible values of  `01/01/24 13:23:00 -  01/01/24 13:23:59`.

asuresh-code: You're correct in that it's not unexpected behaviour for a date range working with date storing seconds. However, the user only sees dates displayed to the minute (as shown below), so they wouldn't know dates are stored to the second, and therefore would not think to extend the range by a minute to capture all possible values.
image
Removing the seconds from the data in the accessor functions shouldn't harm the user experience in anyway, since they can only filter down to the minute.

src/admin/usageStatuses/usageStatuses.component.tsx
asuresh-code: It provides some useful filtering, albeit limited for text values.

If a user wanted to filter for Catalogue Items (Cameras 10 to Cameras 20) precisely, they could with `betweenInclusive`. `fuzzy `and `startsWith `would include other Cameras as well in the results.

You are less likely to find a use case like that for Usage Status values, however I included it anyway since I thought all text columns having the same filter functions throughout would make sense. I do recognise that the use for `betweenInclusive `is limited for text values, so I don't mind removing it from the list.

joshuadkitenge: The mock data used for unit tests is not reflective of data the actually users would put in. So can you remove and don't add to any of the other text value columns. The only text value column this would be applicable for is "serial number" in the items table

src/admin/units/units.component.tsx
joshuadkitenge: This is not a issue for use to solve. The user should use the data range correctly. If they wanted items at 10:07 the range should be 10:06 to 10:08.

joelvdavies: Admittedly I am biased here as I suggested it as an issue 😄.  But I agree with Anikesh’s point – a user deciding what to filter by are given a false impression of the behaviour which I believe is our problem to solve. However as we spoke yesterday about this, there another flaw in that the minimum field would really need to be taken as 0 seconds, and the maximum as 59 seconds for the between filters to work like this rather than just setting the seconds to 0 (e.g. a between filter of 10:07 and 10:08 would exclude 10:07:30 - it would only effect between inclusive), so I think it would be best a separate task anyway if we have between inclusive included.

asuresh-code: I agree that it probably should be left for a seperate task. I spoke to Jousha about this earlier today, and suggested making a request/leaving a comment to MRT for them to fix the issue. 

I will remove the code from this PR. In regards to raising the issue with MRT's creators, do I just make a bug report on their repo, or do something else?

joelvdavies: Yes you would do that, I guess it could be considered a bug. You should also check if there are any similar issues first I did find [this](https://github.com/KevinVandy/material-react-table/issues/993) one that seems related although they just said this up to the implementation to deal with the data first like you did so I am not sure they would address anyway.

joshuadkitenge (approved):
LGTM

joelvdavies (approved):

PR #958, opened 2024-09-11
Catalogue item dialog rhf and zod #946

### LINKED ISSUE #946: Catalogue item dialog RHF and ZOD

joshuadkitenge:
## Description

Refactor the catalogue item dialog so that it uses react hook form and zod 

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #946


src/form.schemas.tsx
```diff
@@ -352,3 +373,45 @@ export const CatalogueCategorySchema = z.object({
     .superRefine((data, ctx) => checkForDuplicatePropertyNames(data, ctx))
     .optional(),
 });
+
+// ------------------------------------ CATALOGUE ITEMS ------------------------------------
+
+export const CatalogueItemSchemaStep1 = (requestType: RequestType) => {
```
joelvdavies, line 379:
```suggestion
export const CatalogueItemDetailsSchema = (requestType: RequestType) => {
```

For consistency since the step 2 is currently `PropertiesStepSchema`.

src/app.types.tsx
```diff
@@ -66,6 +67,37 @@ export interface MoveToCatalogueCategory {
   targetCategory: CatalogueCategory | null;
 }
 
+export interface CatalogueItemStep1 {
```
joelvdavies, line 70:
```suggestion
export interface CatalogueItemDetailsStep {
```

src/app.types.tsx
```diff
@@ -66,6 +67,37 @@ export interface MoveToCatalogueCategory {
...
+  cost_to_rework_gbp?: string | null;
+  days_to_replace: string;
+  days_to_rework?: string | null;
+  drawing_number?: string | null;
+  drawing_link?: string | null;
+  item_model_number?: string | null;
+  notes?: string | null;
+}
+
+export interface CatalogueItemStep1Post {
```
joelvdavies, line 84:
```suggestion
export interface CatalogueItemDetailsStepPost {
```

src/catalogue/items/catalogueItemsDialog.component.tsx
```diff
@@ -30,569 +31,404 @@ import {
...
-    const updatedPropertyErrors = [...propertyErrors];
-    updatedPropertyErrors[index] = false;
-    setPropertyErrors(updatedPropertyErrors);
-    setFormError(false);
-    setFormErrorMessage(undefined);
-  };
+  // Clears form errors when a value has been changed
+  React.useEffect(() => {
+    const subscription1 = watchStep1(() => clearErrorsStep2('root.formError'));
+
```
joelvdavies, line 229:
```suggestion
```

joelvdavies:
Also noticed that clicking edit and removing a manufacturer does this when saving
image
(Clear button or using backspace)

joelvdavies:
> Also noticed that clicking edit and removing a manufacturer does this when saving image (Clear button or using backspace)

This is still the case with backspace. I'm not sure disabling the clear button should be the fix.

joshuadkitenge:
> > Also noticed that clicking edit and removing a manufacturer does this when saving image (Clear button or using backspace)
> 
> This is still the case with backspace. I'm not sure disabling the clear button should be the fix.


https://github.com/user-attachments/assets/aec56720-11bb-41f6-9fc5-d304c01b9e71

joelvdavies (approved):

asuresh-code (approved):
LGTM, Just have 1 comment.
I don't know if it should be addressed in a separate issue, but it would be nice to have some validation for negative numbers where they shouldn't be allowed. Could add an extra "Allowed Values" option like "Range" for Numbers properties. 

image
image

joelvdavies:
> LGTM, Just have 1 comment. I don't know if it should be addressed in a separate issue, but it would be nice to have some validation for negative numbers where they shouldn't be allowed. Could add an extra "Allowed Values" option like "Range" for Numbers properties.
> 
> image image

Both great suggestions, in fact the first one on negative numbers is part of the reason we thought using Zod for schema validation would be useful as it allows that sort of validation to defined as part of the types rather having to add it everywhere. I think we should do that at some point after this is complete. Also the allowed values should hopefully have been written in a way to allow adding another type, I believe we raised it as a possibility but they didn't require it at the time.

joshuadkitenge:
> LGTM, Just have 1 comment. I don't know if it should be addressed in a separate issue, but it would be nice to have some validation for negative numbers where they shouldn't be allowed. Could add an extra "Allowed Values" option like "Range" for Numbers properties.
> 


I have created the issue for the extra validation 

#970

PR #1021, opened 2024-10-09
Redesign landing page items #1019

### LINKED ISSUE #1019: Redesign for items landing page
- [x] The tabs 
- [x] details panel 
- [x] more notes to separate tab

joshuadkitenge:
## Description

see #1019

- Add placeholder image
- Add tabs  
- aad image on details panel

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #1019

### DIFF
--- a/cypress/e2e/with_mock_data/catalogueCategories.cy.ts
+++ b/cypress/e2e/with_mock_data/catalogueCategories.cy.ts
@@ -29,7 +29,7 @@ describe('Catalogue Category', () => {
 
     cy.findByRole('link', { name: 'Motion' }).click();
     cy.findByRole('link', { name: 'Motion' }).should('not.exist');
-    cy.findByText('Actuators').should('not.exist');
+    cy.findByText('Actuators').should('exist');
     cy.findByText('Motion').should('be.visible');
     cy.url().should('include', '/catalogue/2');
   });
--- a/cypress/e2e/with_mock_data/items.cy.ts
+++ b/cypress/e2e/with_mock_data/items.cy.ts
@@ -403,7 +403,7 @@ describe('Items', () => {
     cy.findByText('Zf7P8Qu8TD8c').should('exist');
   });
 
-  it('navigates to the landing page, toggles the properties and navigates back to the table view', () => {
+  it('navigates to the landing page and navigates back to the table view', () => {
     cy.findByText('5YUQDDjKpz2z').click();
     cy.findByText(
       'High-resolution cameras for beam characterization. 1'
--- a/cypress/e2e/with_mock_data/items.cy.ts
+++ b/cypress/e2e/with_mock_data/items.cy.ts
@@ -408,25 +408,6 @@ describe('Items', () => {
     cy.findByText(
       'High-resolution cameras for beam characterization. 1'
     ).should('exist');
-    cy.findByLabelText('Close item properties').should('exist');
-
-    cy.findByLabelText('Close item properties').click();
-
-    cy.findByLabelText('Close item properties').should('not.exist');
-    cy.findByLabelText('Show item properties').should('exist');
-
-    cy.findByLabelText('Close item manufacturer details').should('exist');
-
-    cy.findByLabelText('Close item manufacturer details').click();
-
-    cy.findByLabelText('Close item manufacturer details').should('not.exist');
-
-    cy.findByLabelText('Close item details').should('exist');
-
-    cy.findByLabelText('Close item details').click();
-
-    cy.findByLabelText('Close item details').should('not.exist');
-    cy.findByLabelText('Show item manufacturer details').should('exist');
 
     cy.findByRole('link', {
       name: 'Items',
--- a/src/common/placeholderImage.component.test.tsx
+++ b/src/common/placeholderImage.component.test.tsx
@@ -6,7 +6,7 @@ describe('PlaceholderImage Component', () => {
     const { asFragment } = render(
       <PlaceholderImage maxWidth="300px" maxHeight="200px" />
     );
-    // Take a snapshot of the rendered component
+
     expect(asFragment()).toMatchSnapshot();
   });
 });
--- a/src/common/placeholderImage.component.test.tsx
+++ b/src/common/placeholderImage.component.test.tsx
@@ -1,0 +1,12 @@
+import { render } from '@testing-library/react';
+import PlaceholderImage from './placeholderImage.component'; // Adjust the import path as necessary
+
+describe('PlaceholderImage Component', () => {
+  it('matches the snapshot', () => {
+    const { asFragment } = render(
+      <PlaceholderImage maxWidth="300px" maxHeight="200px" />
+    );
+    // Take a snapshot of the rendered component
+    expect(asFragment()).toMatchSnapshot();
+  });
+});
--- a/src/common/placeholderImage.component.tsx
+++ b/src/common/placeholderImage.component.tsx
@@ -15,12 +15,12 @@ const PlaceholderImage = (props: PlaceholderImageProps) => {
         width: '100%',
         height: '100%',
         borderRadius: 2,
-        backgroundColor: 'inherit', // Light gray background
+        backgroundColor: 'inherit',
         display: 'flex',
         alignItems: 'center',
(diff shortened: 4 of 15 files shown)


joshuadkitenge:
image

asuresh-code (approved):
LGTM, just wanted to check that the "placeholder_image" was meant to be the "No Image" icon, and not something like this.
image  image

joshuadkitenge:
> LGTM, just wanted to check that the "placeholder_image" was meant to be the "No Image" icon, and not something like this. image image

Yeah it is mean to be like this. I choose this style because it work best with the different theme modes when in scigateway

joelvdavies:
Just a quick thing before I look at the code - as I mentioned before I think it feels a bit tight between the details and the next heading I would prefer double spacing after a section or something e.g.

image
instead of
image

Just to make it visually a bit more separate as though they are separate paragraphs.

Also could the actions section move over to the left inline with the ends of the dividers (seems it just needs a margin-left: auto when I tried it).

src/view/viewTabs.component.tsx
```diff
@@ -102,7 +71,7 @@ function ViewTabs() {
               height: 'calc(100% - 48px)',
             }}
           >
-            <TabPanel value={value} label={value}>
+            <TabPanel<TabValue> value={value} label={value}>
```
joelvdavies, line 74:
```suggestion
            <TabPanel value={value} label={value}>
```
It will automatically handle the type because it already has the type from the type of value. I found it a bit confusing at first and couldn't find examples of anyone else using templates like that in jsx. Same for the others.

src/utils.tsx
```diff
@@ -421,3 +424,42 @@ export const checkForDuplicates = (props: {
...
+    id: `${label}-tab`,
+    'aria-controls': `${label}-tabpanel`,
+  };
+}
+
+export const StyledTab = styled(Tab)(({ theme }) => ({
+  textTransform: 'none',
+  fontWeight: theme.typography.fontWeightBold,
+  fontSize: theme.typography.pxToRem(16),
+}));
```
joelvdavies, line 465: This isn't really a utility function at this point, I think its better in common as a component instead.

src/items/itemsLandingPage.component.tsx
```diff
@@ -24,10 +29,32 @@ import { useGetItem } from '../api/items';
...
+] as const;
+
+const iconMapping: Record<ItemsTabValue, React.ReactElement> = {
+  Information: <InfoOutlinedIcon />,
+  Gallery: <CollectionsOutlinedIcon />,
+  Attachments: <AttachmentOutlinedIcon />,
+  Notes: <NotesIcon />,
+};
+
+type ItemsTabValue = (typeof ITEMS_TAB_VALUES)[number];
```
joelvdavies, line 56:
```suggestion
const ITEMS_TAB_VALUES = [
  'Information',
  'Gallery',
  'Attachments',
  'Notes',
] as const;

type ItemsTabValue = (typeof ITEMS_TAB_VALUES)[number];

const iconMapping: Record<ItemsTabValue, React.ReactElement> = {
  Information: <InfoOutlinedIcon />,
  Gallery: <CollectionsOutlinedIcon />,
  Attachments: <AttachmentOutlinedIcon />,
  Notes: <NotesIcon />,
};
```
So that its in the order its used.

src/common/placeholderImage.component.test.tsx
```diff
@@ -0,0 +1,12 @@
+import { render } from '@testing-library/react';
+import PlaceholderImage from './placeholderImage.component'; // Adjust the import path as necessary
+
+describe('PlaceholderImage Component', () => {
+  it('matches the snapshot', () => {
+    const { asFragment } = render(
+      <PlaceholderImage maxWidth="300px" maxHeight="200px" />
+    );
+    // Take a snapshot of the rendered component
```
joelvdavies, line 9:
```suggestion
```

src/common/placeholderImage.component.tsx
```diff
@@ -0,0 +1,33 @@
...
+        maxHeight: maxHeight,
+        width: '100%',
+        height: '100%',
+        borderRadius: 2,
+        backgroundColor: 'inherit', // Light gray background
+        display: 'flex',
+        alignItems: 'center',
+        justifyContent: 'center',
+        textAlign: 'center', // Center the text
+        color: theme.palette.text.primary, // Text color
```
joelvdavies, line 23: Are these comments yours?

cypress/e2e/with_mock_data/items.cy.ts
```diff
@@ -408,25 +408,6 @@ describe('Items', () => {
     cy.findByText(
       'High-resolution cameras for beam characterization. 1'
     ).should('exist');
```
joelvdavies, line 410: The name of this test should change to reflect these changes.

cypress/e2e/with_mock_data/catalogueCategories.cy.ts
```diff
@@ -29,7 +29,7 @@ describe('Catalogue Category', () => {
 
     cy.findByRole('link', { name: 'Motion' }).click();
     cy.findByRole('link', { name: 'Motion' }).should('not.exist');
-    cy.findByText('Actuators').should('not.exist');
+    cy.findByText('Actuators').should('exist');
```
joelvdavies, line 32: I don't think the assertion here makes sense anymore. It was checking Actuators is visible, then clicking a link and checking it doesn't exist anymore. Where as now its asserting it is there both before and after - so maybe it only needs the Motion one?

joelvdavies (changes requested):
Works well, I like the design. Just a few minor comments.

src/view/viewTabs.component.tsx
joshuadkitenge: I did this for consistency. i know I didn't need to put it here but I've used if for the other ones

joelvdavies: Could you just remove them then? I struggled to find anyone else using it and chatgpt suggested its best practice to leave it out for this sort of case.

cypress/e2e/with_mock_data/catalogueCategories.cy.ts
joshuadkitenge: there is no harm in including this test. checking for actuators to exist. tells us that the page has fully loaded.

src/common/placeholderImage.component.tsx
joshuadkitenge: I'll remove them

cypress/e2e/with_mock_data/catalogueCategories.cy.ts
joelvdavies: True but it doesn't actually do anything now considering it already has a check for Motion not existing as  a link but being visible elsewhere.

joshuadkitenge:
> Just a quick thing before I look at the code - as I mentioned before I think it feels a bit tight between the details and the next heading I would prefer double spacing after a section or something e.g.
> 
> image instead of image
> 
> Just to make it visually a bit more separate as though they are separate paragraphs.
> 
> Also could the actions section move over to the left inline with the ends of the dividers (seems it just needs a `margin-left: auto` when I tried it).

that has been done in the next pr (action button) #1022

src/view/viewTabs.component.tsx
joshuadkitenge: I add it for consistency within our codebase as this is how it is use in other places. This makes it easy to maintain in the future

joelvdavies: Although now I actually have the proper name, I can find others using it was just that it was relatively new.

joelvdavies (approved):

asuresh-code (approved):

src/view/viewTabs.component.tsx
joelvdavies: I also just realised there are other places we don’t specify the generic types in JSX (not that I want us to 😄 ) e.g. <MaterialReactTable table={table} />, a lot of MUI components ListItemText, TextField etc, nobody uses it though because there is no need.

PR #1046, opened 2024-10-16
Enhance Items Table Filtering

### LINKED ISSUE #1041: Enhance Items Table Filtering

asuresh-code:
## Description

Serial Number:

- Filter variant = Text
_Data is string which leaves 5 options: select, checkbox, autocomplete, multiselect and text. Text and Autocomplete offer the most filtering functionality. Every Record has a unique serial number, so autocomplete is not necessary._
- Filter mode = [Fuzzy, Contains, Starts With, Ends With, Equals, Not Equals, Between Inclusive]
_The first 6 are the standard filter functions for text (as used in the previous PR). Between Inclusive was added just for this text column since it allows useful filtering for some patterns (i.e SFDE12 - SFDE 23)_

Last Modified/Created:
- Filter variant = Datetime-Range
_Datetime is chosen since the data has that level of specificity. Range is chosen since it allows more functionality, and filtering for an exact datetime is not needed._
- Filter mode = ['between inclusive', 'between']:
_The standard filter functions for date-time ranges, following on from the previous table._

Warranty End Date/Delivered Date:
- Filter Variant = Date-Range
_These dates are displayed without the time, so date is the defacto variant._
- Filter Mode = ['between inclusive', 'between', 'equals', 'not equals]
_Between (inclusive) are the standard functions for range variants. Since the date does not include time, the equals/not equals functions can be useful (i.e filter for a specific delivery day) and have been included._

Asset Number/Purchase Order Number/Notes:
- Filter Variant = Text
_The defacto variant for string data; text and autocomplete provide the most functionality and the data is unique for each record, thus text is used._
- Filter mode = [Fuzzy, Contains, Starts With, Ends With, Equals, Not Equals, Empty, Not Empty]
_The standard filtering modes for string data, as used in the previous table PR. Also they are all optional fields, and so have empty/not empty filters as well._

Usage Status/Systems:
- Filter Variant = Multi-Select
_Users asked for a filter variant to filter by multi-selection_
- Filter Mode = [arrIncludesSome, arrIncludesNone (custom)]
_Users asked to be able to filter by inclusion and exclusion, so these 2 functions provide that functionality_

Is Defective:
- Filter Variant = Select
_The data is boolean, and select provides the variant to easily filter by true or false. Checkbox offers similar functionality, but is less appealing to use._
- Filter Mode = N/A, no other functions needed

Properties:
- Filter Variant: Boolean = Select, String = Text, Number = Text, Null = Text
_This uses the existing filter variants, except for boolean. Again, the checkbox variant is built specifically for booleans so was selected. Text is used as the filter variant for numbers since it allows for more functions to be used (equals, not equals), whilst providing the same functionality as range for a filter variant._
- Filter Mode: Boolean = [] , String | Null =  [Fuzzy, Contains, Starts With, Ends With, Equals, Not Equals], Number = ['between inclusive', 'between', 'equals', 'not equals,]
_These are the standard filter modes used previously for string values, and number (date) values respectively. Also if a property is not mandatory, empty/not empty is added._

**N.B: Optional Dates do not have the empty/not empty filter functions, because MRT's filter variants do not seem to be compatible with them (as shown below)**
Broken empty date filter

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

Closes #1041


asuresh-code:
> _Usage Status/Systems:
> 
> ```
> Filter Variant = Multi-Select
> Users asked for a filter variant to filter by multi-selection
> Filter Mode = [arrIncludesSome, arrIncludesNone (custom)]
> Users asked to be able to filter by inclusion and exclusion, so these 2 functions provide that functionality_
> ```
> 
> for this section are the custom function needed can't you use equals and not equals ?

Unfortunately not, When I tried using equals, and not equals with the multi-select filter, it seems to compare the whole filter array with the record value. So with equals, if you select 2 or more filter values no records appear, and with not equals, every record appears.

src/utils.tsx
```diff
@@ -413,6 +413,23 @@ export const getInitialColumnFilterFnState = <TData extends MRT_RowData>(
...
 
+interface FilterFn {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (row: MRT_RowData, id: string, filterValue: any): boolean;
+}
+
+export const customFilterFunctions: Record<string, FilterFn> = {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  arrIncludesNone: (row: MRT_RowData, id: string, filterValue: any) => {
+    return !filterValue.includes(row.getValue(id));
```
joelvdavies, line 424:
```suggestion
  (row: MRT_RowData, id: string, filterValue: string | number): boolean;
}

export const customFilterFunctions: Record<string, FilterFn> = {
  arrIncludesNone: (row: MRT_RowData, id: string, filterValue: string | number) => {
    return !filterValue.includes(row.getValue(id));
```
The type seems to be on https://www.material-react-table.com/docs/guides/column-filtering.

joelvdavies:
Could you put `Closes` instead of `connects to` in the PR please so it auto closes the issue when merged. (Same for https://github.com/ral-facilities/inventory-management-system-api/pull/393)

joshuadkitenge (changes requested):
For the isDefective and other boolean value columns, please keep the select component instead of using checkboxes, as the checkboxes do not look appealing.

Also, increase the width of the multi-select columns.

asuresh-code:
I didn't consider this before, but would it be worth adding `empty `and `not empty` filters (inbuilt MRT ones) for any optional fields? @joshuadkitenge @joelvdavies

joshuadkitenge:
> I didn't consider this before, but would it be worth adding `empty `and `not empty` filters (inbuilt MRT ones) for any optional fields? @joshuadkitenge @joelvdavies

yeah

asuresh-code:
> > I didn't consider this before, but would it be worth adding `empty `and `not empty` filters (inbuilt MRT ones) for any optional fields? @joshuadkitenge @joelvdavies
> 
> yeah

I've been able to add empty and not empty to all of the optional columns, except for date columns. The column sort of breaks (I've attached an image in the PR description

src/items/itemsTable.component.tsx
```diff
@@ -219,23 +239,58 @@ export function ItemsTable(props: ItemTableProps) {
...
-        filterVariant: 'autocomplete',
+        filterVariant: 'multi-select',
+        filterFn: 'arrIncludesSome',
+        columnFilterModeOptions: ['arrIncludesSome', 'arrIncludesNone'],
+        renderColumnFilterModeMenuItems: ({ onSelectFilterMode }) => [
+          <MenuItem
+            key="arrIncludesSome"
+            onClick={() => onSelectFilterMode('arrIncludesSome')}
+          >
+            Includes
```
joshuadkitenge, line 260: Can you add units test any additional logic you added that wasn't built into MRT.  i.e "customFilterFunctions"

src/items/itemsTable.component.tsx
```diff
@@ -163,16 +168,28 @@ export function ItemsTable(props: ItemTableProps) {
...
         size: 250,
         GroupedCell: TableGroupedCell,
       },
       {
         header: 'Purchase Order Number',
         Header: TableHeaderOverflowTip,
-        accessorFn: (row) => row.item.purchase_order_number,
+        accessorFn: (row) => row.item.purchase_order_number ?? '',
         id: 'item.purchase_order_number',
+        filterVariant: COLUMN_FILTER_VARIANTS['string'],
```
joelvdavies, line 187: I like the reuse here. I would personally prefer using `.string` over the array notation as while it can be accessed like an array it isn't really and I think its slightly neater. Also in two minds on the capitalisation, I think I have typically seen such things like `ColumnFilterVariants.STRING` over ` COLUMN_FILTER_VARIANTS.string` e.g. https://stackoverflow.com/questions/838929/naming-why-should-named-constants-be-all-uppercase-in-c-java.

asuresh-code: Will implement the dot notation. With the capitalisation, is the rationale for the former that the actual constant value in use is the value, not the whole data structure? Therefore, capitalise the key, not the array/dictionary/etc?

joelvdavies: Yes I believe so. The ColumnFilterVariants is an object containing the constants rather than a constant itself.

joelvdavies: But then the google style guide currently has https://google.github.io/styleguide/tsguide.html#identifiers-constants. So it looks like the original you had before is preferred after all 😅

asuresh-code: Ah, it's no problem😅. I didn't even know the google style guide was a resource for styling conventions, so now I know more!

joelvdavies: Yeah we don't necessarily conform to a specific style guide nor have one ourselves, we decide what we consider to be best practice based on what others do. This is one of those cases, since google uses that it seems sensible to follow it.

src/items/itemsTable.component.tsx
joshuadkitenge: No, high code coverage alone doesn’t guarantee thorough testing, as it only shows that lines were executed, not that all cases were covered.

src/items/itemsTable.component.tsx
```diff
@@ -264,70 +323,65 @@ export function ItemsTable(props: ItemTableProps) {
...
-              ? 'Yes'
-              : 'No';
+            const booleanPropertyValue: boolean = propertyValue as boolean;
+            if (booleanPropertyValue === true) {
+              return 'Yes';
+            } else if (booleanPropertyValue === false) {
+              return 'No';
+            } else {
+              return '';
+            }
```
joshuadkitenge, line 339:
```suggestion
      if (typeof propertyValue === 'boolean') {
        return propertyValue ? 'Yes' : 'No';
      } else {
        return '';
      }
```

src/utils.test.tsx
```diff
@@ -318,6 +319,41 @@ describe('Utility functions', () => {
...
+    it('should correctly exclude record, when filter value is not a list', () => {
+      const result = filterExclude(person, 'status', 'unemployed');
+      expect(result).toBe(false);
+    });
+    it('should correctly include record, when filter value is not a list', () => {
+      const result = filterExclude(person, 'age', 3);
+      expect(result).toBe(true);
+    });
+  });
+});
```
joshuadkitenge, line 355:
```suggestion
describe('customFilterFunctions', () => {
  describe('arrIncludesNone', () => {
    const person: MRT_RowData = {
      name: 'Dan',
      age: 4,
      status: 'unemployed',
      getValue: (id: string) => {
        return person[id as keyof MRT_RowData]; // Return the corresponding field from the object
      },
    };
    const filterExclude: (
      row: MRT_RowData,
      id: string,
      // eslint-disable-next-line @typescript-eslint/no-explicit-any
      filterValue: any
    ) => boolean = customFilterFunctions['arrIncludesNone'];

    it('should correctly exclude record', () => {
      const result = filterExclude(person, 'status', ['unemployed']);
      expect(result).toBe(false);
    });
    it('should correctly include record', () => {
      const result = filterExclude(person, 'age', [8, 29]);
      expect(result).toBe(true);
    });
    it('should correctly exclude record, when filter value is not a list', () => {
      const result = filterExclude(person, 'status', 'unemployed');
      expect(result).toBe(false);
    });
    it('should correctly include record, when filter value is not a list', () => {
      const result = filterExclude(person, 'age', 3);
      expect(result).toBe(true);
    });
  });
});
```

The beforeAll block isn't necessary if you don’t need initializing person and filterExclude directly within each test.

src/utils.tsx
```diff
@@ -413,6 +414,77 @@ export const getInitialColumnFilterFnState = <TData extends MRT_RowData>(
...
+    }
+    return row.getValue(id) !== filterValue;
+  },
+};
+
+export const MRT_Functions_Localisation: Record<string, string> = {
+  filterArrIncludesNone: 'Excludes',
+};
+
+export type PropertyFiltersType = {
```
joshuadkitenge, line 436: Now that this has been moved to utils, the name PropertyFiltersType could be confusing. It might help to choose a more descriptive name that reflects its purpose and scope within utils.

src/items/itemsTable.component.tsx
```diff
@@ -219,23 +238,57 @@ export function ItemsTable(props: ItemTableProps) {
...
+            Includes
+          </MenuItem>,
+          <MenuItem
+            key="arrIncludesNone"
+            onClick={() => onSelectFilterMode('arrIncludesNone')}
+          >
+            Excludes
+          </MenuItem>,
+        ],
+        size: 250,
```
joshuadkitenge, line 267: Can you increase the size of the usage statuses column to the same as the dates please. Also can look at all the column sizes to make sure that they are appropriate. The purchase order number seems to be too big.

joshuadkitenge: Can you also increase the system column as well

joshuadkitenge (changes requested):
Looks good overall, just a few minor comments.

Since you've updated the properties logic in the table, this PR will be blocked until PR #1040 is ready for review. Once reviewed, the PRs can be merged together.

src/utils.tsx
joelvdavies: Not sure what a good name would be, but the new one of `ItemPropertyFiltersTypes` feels more confusing to me. It applies to items and catalogue items which is why Item wasn't mentioned in the name before. If its used in `ColumnFiltersType` perhaps `PropetyColumnFiltersType` is more consistent?

asuresh-code: It's only used in the context of item properties in columns, so I think calling it `PropertyColumnFiltersType` is better and works with its use case. Yeah, I forgot that it was used in both Catalogue Items and Items table, so I agree omitting `Item` is probably best.

I think maybe a name like `PropertyFilterVariantsType` and `ColumnFilterVariantsType` would be better? Makes it clearer that it's for setting filter variants, and not something like filter functions.

joshuadkitenge: Is this even needed anymore ??

asuresh-code: Do you mean using ColumnFiltersType everywhere instead of the more specific PropertyColumnFiltersType? You could replace it in the propertyFilters definition, you would just have to add fields for date, and datetime in the instantiation.

```
const propertyFilters: ColumnFiltersType = {
      boolean: 'autocomplete',
      string: 'text',
      number: 'range',
      null: 'text',
      date: 'date',
      datetime: 'datetime',
    };
```

joshuadkitenge: The `propertyFilters` for the item table is only used here 

```
export type ColumnFiltersType = PropertyFiltersType & {
  datetime: 'datetime-range' | 'datetime';
  date: 'date-range' | 'date';
};
```
it can be moved into this.

asuresh-code: Oh, yes I see what you mean. Yeah, it's not really being used individually in the items Table. The only reason I've kept it is because it's still used in Catalogue Items table, but once I've reworked that table, it'll no longer be needed.

joelvdavies (approved):

PR #1151, opened 2024-12-11
Edit Dialog for Images

### LINKED ISSUE #1075: Edit images
#### Tasks:
- [ ] Create a **dialog** for editing the metadata of images (e.g., title, description). 
- [ ] Add an **Edit** `MenuItem` to the **Images** for each image in the tile action menu. 
- [ ] Implement a **PATCH** endpoint to update images metadata in the backend.

#### notes:
might be able to reuse this dialog #1068

asuresh-code:
## Description

see #1075 

**N.B. Needs to wait for [corresponding backend PR ](https://github.com/ral-facilities/object-storage-api/pull/75)to merge.**

## Testing instructions

**Checkout `**add-update-endpoint-without-primary-for-images-#35**` on the Object Storage API repo to test with the api.**

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #1075


src/common/images/editFileDialog.component.tsx
```diff
@@ -0,0 +1,196 @@
...
+  open: boolean;
+  onClose: () => void;
+  selectedFile?: APIImage;
+  fileType: 'Image' | 'Attachment';
+  usePatchFile: () => UseMutationResult<
+    APIImage,
+    AxiosError,
+    { id: string; fileMetadata: ObjectFilePatch }
+  >;
+}
```
joshuadkitenge, line 34: Can use use a discrimination union for the type. to distinguish between image and attachments type  
https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html

joshuadkitenge:
Can you amove the editFileDialog out of common/images to common/ please

cypress/e2e/with_mock_data/items.cy.ts
```diff
@@ -894,7 +894,92 @@ describe('Items', () => {
...
+        .within(() => {
+          cy.findByRole('button', { name: 'Save' }).click();
+          cy.contains(
+            "There have been no changes made. Please change a field's value or press Cancel to exit."
+          );
+        });
+      cy.findByRole('button', { name: 'Save' }).should('be.disabled');
+    });
+
+    it('Required fields that are cleared are not allowed and show error message', () => {
```
rowan04, line 960:
```suggestion
    it('shows error message when required fields are cleared', () => {
```

rowan04 (commented):
just some grammar rewriting for the if test descriptions. The rest of the code looks good I think but I haven't tested the functionality.

joshuadkitenge (changes requested):

rowan04:
why does one of the images not show for me? it seems to happen for any catalogue item, always the 3rd image. I assume its not to do with your changes here but maybe you know why
image

joshuadkitenge:
> why does one of the images not show for me? it seems to happen for any catalogue item, always the 3rd image. I assume its not to do with your changes here but maybe you know why image

That is on purpose for the unit and e2e test with the mock data

joshuadkitenge:
Can you add the end adornment to the uppy upload dialog file in this pr please

cypress/e2e/with_mock_data/items.cy.ts
```diff
@@ -930,12 +931,12 @@ describe('Items', () => {
...
         const request = patchRequests[0];
         expect(JSON.stringify(await request.json())).equal(
-          '{"file_name":"test file","description":"test description","title":"test title"}'
+          '{"file_name":"test file.png","description":"test description","title":"test title"}'
         );
       });
     });
 
-    it('not changing any fields shows error', () => {
+    it('shows error when no fields have been changed', () => {
```
rowan04, line 939:
```suggestion
    it('shows error message when no fields have been changed', () => {
```
apologies i know i changed it to this the first time :)

joshuadkitenge:
> Can you add the end adornment to the uppy upload dialog file in this pr please

Can this be done for both dialog images and attachments

joshuadkitenge (changes requested):

asuresh-code:
- Currently the `uppy.utils.tsx` does not have a test file. Is this something that needs to be changed?
- I'm unable to view the codecov report to see what lines are missing; Is this an issue others are having?
- The implementation for checking if dark mode is enabled is the same as what SciGateway does. I tried using the light-dark implemntation seen [here](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/light-dark), but it did not work (I'm not sure why).

joshuadkitenge:
> * Currently the `uppy.utils.tsx` does not have a test file. Is this something that needs to be changed?
> 
>     * I'm unable to view the codecov report to see what lines are missing; Is this an issue others are having?
> 
>     * The implementation for checking if dark mode is enabled is the same as what SciGateway does. I tried using the light-dark implemntation seen [here](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/light-dark), but it did not work (I'm not sure why).

3. Look a the third button point for how to check if it ligh or dark mode https://github.com/ral-facilities/inventory-management-system/pull/1151#discussion_r1923969268

joshuadkitenge (changes requested):
> * Currently the `uppy.utils.tsx` does not have a test file. Is this something that needs to be changed?
> 
>     * I'm unable to view the codecov report to see what lines are missing; Is this an issue others are having?
> 
>     * The implementation for checking if dark mode is enabled is the same as what SciGateway does. I tried using the light-dark implemntation seen [here](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/light-dark), but it did not work (I'm not sure why).
1. The contents of uppy.utils.tsx do not need direct testing, as they are indirectly tested through other components or functions that utilize this file. The current contents of this file is never used as a standalone module, so testing its integration with the uppy is what’s important. However, this statement applies only to the current contents of the file. Any new functionality added in the future should be evaluated to determine if direct testing is necessary.
2. i had the same issue logging into codecov solved the issue for me. You should also be able to see the report local when you run the tests  
3. Look a the third button point for how to check if it light or dark mode https://github.com/ral-facilities/inventory-management-system/pull/1151#discussion_r1923969268.
3.1.  Do the same implemenation as uppy does.

src/common/images/uploadImagesDialog.component.tsx
```diff
@@ -109,6 +112,69 @@ const UploadImagesDialog = (props: UploadImagesDialogProps) => {
...
+                      colorScheme: 'light',
+                      height: '31px',
+                      padding: '5px',
+                    },
+                  },
+                  '.' + extension
+                ),
+              ]
+            );
+          },
```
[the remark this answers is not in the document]
joshuadkitenge: > * As the metafields for attachements and images are the same can you make a general object to set them both

you have current done this for just the name this should be done the whole metafields

asuresh-code:
> 3. Look a the third button point for how to check if it light or dark mode [Edit Dialog for Images #1151 (comment)](https://github.com/ral-facilities/inventory-management-system/pull/1151#discussion_r1923969268).
>    3.1.  Do the same implemenation as uppy does.

I can see that uppy creates a CSS class that conditionally sets the style based on a variable data-uppy-theme
image
I could copy the CSS class they use and style the div element with that when the input element is focussed, which would mean I wouldn't need the condition to check if it's in dark mode. However this would require a separate CSS file in our repo. I can't use the exact same styling (import style from uppy somehow), because it would only apply when div is focussed, whilst I need it to apply when input is focussed.

 _I did try and see if there was any way to trick the div element's CSS into think its being focussed, without actually focussing on it, but I couldn't find anything that worked_

_CSS also has the `focus-within` property which would be perfect for this case, since the input elem being triggered would cause any styling on the div with this as a condition to be applied, but again this would require changing the CSS class, which I don't think you can do using inline CSS, so it would also require a separate CSS stylesheet in our repo._

I tried passing the theme prop from the dashboard to the render function, however it does not update when I switch themes.

I think if I want to do uppy's implementation I would need to either: a. create a separate CSS file to create a CSS class in the same way uppy does, or b. use data-uppy-theme as the condition on whether it is dark (however it gets this from theme prop, which I also passed to my function, but it does not mutate).

joshuadkitenge:
> > 3. Look a the third button point for how to check if it light or dark mode [Edit Dialog for Images #1151 (comment)](https://github.com/ral-facilities/inventory-management-system/pull/1151#discussion_r1923969268).
> >    3.1.  Do the same implemenation as uppy does.
> 
> I can see that uppy creates a CSS class that conditionally sets the style based on a variable data-uppy-theme image I could copy the CSS class they use and style the div element with that when the input element is focussed, which would mean I wouldn't need the condition to check if it's in dark mode. However this would require a separate CSS file in our repo. I can't use the exact same styling (import style from uppy somehow), because it would only apply when div is focussed, whilst I need it to apply when input is focussed.
> 
> _I did try and see if there was any way to trick the div element's CSS into think its being focussed, without actually focussing on it, but I couldn't find anything that worked_
> 
> _CSS also has the `focus-within` property which would be perfect for this case, since the input elem being triggered would cause any styling on the div with this as a condition to be applied, but again this would require changing the CSS class, which I don't think you can do using inline CSS, so it would also require a separate CSS stylesheet in our repo._
> 
> I tried passing the theme prop from the dashboard to the render function, however it does not update when I switch themes.
> 
> I think if I want to do uppy's implementation I would need to either: a. create a separate CSS file to create a CSS class in the same way uppy does, or b. use data-uppy-theme as the condition on whether it is dark (however it gets this from theme prop, which I also passed to my function, but it does not mutate).

Implement option B. Option A doesn’t align well with the coding style we use for defining CSS for elements.

asuresh-code:
> Implement option B. Option A doesn’t align well with the coding style we use for defining CSS for elements.

I've tried implementing this option, however the theme prop in the Render function does not update when the theme changes. It does update in the `uploadDialog` component, but not in the render function. 

I've tried passing it using React.useState and explicitly updating it in the parent component, but the change doesn't trickle down, and I can't call useTheme() to check the theme directly in the function since it is a React Hook and violates React Hook Laws by doing so.

joshuadkitenge:
> > Implement option B. Option A doesn’t align well with the coding style we use for defining CSS for elements.
> 
> I've tried implementing this option, however the theme prop in the Render function does not update when the theme changes. It does update in the `uploadDialog` component, but not in the render function.
> 
> I've tried passing it using React.useState and explicitly updating it in the parent component, but the change doesn't trickle down, and I can't call useTheme() to check the theme directly in the function since it is a React Hook and violates React Hook Laws by doing so.

This should be possible to do

joshuadkitenge:
> * As the metafields for attachements and images are the same can you make a general object to set them both

you have current done this for just the name this should be done the whole metafields @asuresh-code

asuresh-code:
> > * As the metafields for attachements and images are the same can you make a general object to set them both
> 
> you have current done this for just the name this should be done the whole metafields @asuresh-code
> 
> Hint: const { metafields } = useMetaFields();

I'm in the process of writing a function that generalises all of the metafields instead of just the filename field, and just wanted to check if `useMetaFields();` is an example name of the function and not an actual one that exists? I checked the uppy repo, and couldn't find an entry under that name?

src/common/editFileDialog.component.tsx
```diff
@@ -0,0 +1,221 @@
...
+}
+
+export type FileDialogProps = ImageDialogProps;
+
+const EditFileDialog = (props: FileDialogProps) => {
+  const { open, onClose, selectedFile, fileType, usePatchFile } = props;
+
+  const { mutateAsync: patchFile, isPending: isEditPending } = usePatchFile();
+
+  const [initialName, extension] = splitFilename(selectedFile.file_name);
```
joshuadkitenge, line 49:
```suggestion
  const [initialName, extension] = getNameAndExtension(selectedFile.file_name);
```

src/common/attachments/uploadAttachmentsDialog.component.test.tsx
```diff
@@ -87,11 +87,32 @@ describe('Upload attachment dialog', () => {
...
+    expect(await screen.findByText('File name')).toBeInTheDocument();
+
+    // Checks if file extension is displayed. If it's editable, actual value will disappear after editing.
+    expect(await screen.findByText('.txt')).toBeInTheDocument();
+
+    const [name, title, description] = screen.getAllByRole('textbox');
+
+    await user.type(name, 'test.pdf');
+    await user.type(title, 'test title');
+    await user.type(description, 'test description');
```
joshuadkitenge, line 103: This approach using `getAllByRole('textbox')` is not ideal for accessibility. By relying only on the role and index, it becomes difficult to ensure that the correct input field is being interacted with.

src/common/editFileDialog.component.tsx
```diff
@@ -0,0 +1,207 @@
...
+  const {
+    handleSubmit,
+    register,
+    formState: { errors },
+    watch,
+    setError,
+    clearErrors,
+    reset,
+  } = useForm<ObjectFilePatch>({
+    resolver: zodResolver(FileSchemaPatch()),
```
joshuadkitenge, line 67: The FileSchemaPatch() implementation here is unnecessary because FileSchemaPatch doesn't require any parameters or dynamic evaluation. Since it’s not a function or callable that needs to be invoked, it’s better to pass it directly to the resolver as `zodResolver(FileSchemaPatch)`.

src/common/uppy.utils.tsx
```diff
@@ -9,4 +14,131 @@ export const StyledUppyBox = styled(Box)(({ theme }) => ({
     width: 'calc(100% - 64px)',
   },
   '& .uppy-Dashboard--modal .uppy-Dashboard-inner': { zIndex: 1300 },
+  // The theme variable doesn't update as expected in the renderFields function.
+  // Therefore, all CSS that depends on the theme mode is handled here.
+  // The input focus styles are adapted from Uppy core styles:
+  // https://github.com/transloadit/uppy/blob/3aa8d688ff4a94b52287dbf737eb5b3f10b29dda/packages/%40uppy/core/src/_common.scss#L49
+  '.custom-div-focus': {
```
joshuadkitenge, line 21: The CSS class name custom-div-focus is too generic and could lead to potential naming clashes. To make it more descriptive and context-specific, consider including both ims and uppy in the class name, along with its purpose (focus state within an input).

src/common/uppy.utils.tsx
```diff
@@ -9,4 +14,131 @@ export const StyledUppyBox = styled(Box)(({ theme }) => ({
...
+            padding: '5px',
+          },
+        },
+        extension
+      ),
+    ]
+  );
+}
+
+//Uppy custom type: MetaField[] type imported from node modules:
```
joshuadkitenge, line 120:
```suggestion
```
This is not needed.

src/form.schemas.tsx
```diff
@@ -616,3 +616,14 @@ export const ItemDetailsStepSchema = (requestType: RequestType) => {
...
+// ------------------------------------ FILES ------------------------------------
+
+export const FileSchemaPatch = () =>
+  z.object({
+    file_name: MandatoryStringSchema({
+      errorMessage: 'Please enter a file name.',
+    }),
+    title: NullableStringSchema,
+    description: NullableStringSchema,
+  });
```
joshuadkitenge, line 629:
```suggestion
export const FileSchemaPatch = 
  z.object({
    file_name: MandatoryStringSchema({
      errorMessage: 'Please enter a file name.',
    }),
    title: NullableStringSchema,
    description: NullableStringSchema,
  });
```

joshuadkitenge (changes requested):

joshuadkitenge (changes requested):

src/common/images/uploadImagesDialog.component.test.tsx
```diff
@@ -48,6 +58,91 @@ describe('Upload image dialog', () => {
...
+    await waitFor(() => {
+      expect(screen.getByText('image.png')).toBeInTheDocument();
+    });
+
+    await user.click(
+      await screen.findByRole('button', { name: 'Edit file image.png' })
+    );
+
+    expect(await screen.findByText('File name')).toBeInTheDocument();
+    const [filename, title, _] = screen.getAllByRole('textbox');
```
joshuadkitenge, line 89: same as above

joshuadkitenge (approved):

PR #1152, opened 2024-12-11
Uppy refresh token logic

### LINKED ISSUE #1147: Use API refresh logic on image upload
See comment here: https://github.com/ral-facilities/inventory-management-system/pull/1145#discussion_r1871151219. If the token times out while uploading, it wont refresh when uploading images using uppy as that logic is handled by axios which it doesn't use.

I think it would be worth moving this logic into the API, along with any query invalidation when doing this to ensure the logic is kept together.

joshuadkitenge:
## Description

- Fix existing code so it use relevant api client instead of imsApi
- refactor thumbnail view logic to use displayed images
- remove upload error uppy event listener as it not needed as backend deals with image errors
- create uppy uppyOnBeforeRequest function that set the auth headers
- create uppyOnAfterResponse to deal with 403 errors
## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
closes #1147


src/api/api.tsx
```diff
@@ -91,6 +92,66 @@ const createAuthenticatedClient = (props: {
...
+              resolve(); // Resolve the promise when the token is refreshed
+            }
+          };
+
+          const timeoutId = setTimeout(() => {
+            // If the token isn't refreshed within a reasonable timeframe, reject the promise
+            document.removeEventListener(MicroFrontendId, handler);
+            isFetchingAccessToken = false;
+            reject();
+          }, 20 * 1000); // 20 seconds timeout
```
joshuadkitenge, line 137: Added a reject case to handle scenarios where the expected CustomEvent does not occur or takes too long.
Ensures the Promise does not hang indefinitely

joshuadkitenge:
remove timeout

PR #1159, opened 2024-12-16
 Create a common AllowedValuesListTextFields for RHF dialogs #1158

### LINKED ISSUE #1158: Create a common AllowedValuesListTextFields for RHF dialogs

joshuadkitenge:
## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #1158


src/catalogue/category/property/allowedValuesListTextFields.component.tsx
```diff
@@ -0,0 +1,226 @@
...
+  >();
+
+  const propertyRHF = watch();
+
+  const { fields, append, remove } = useFieldArray({
+    control,
+    name:
+      typeof nestIndex === 'number' && 'properties' in propertyRHF
+        ? `properties.${nestIndex}.allowed_values.values.values`
+        : `allowed_values.values.values`, // Adjust the field name according to your data structure
```
joshuadkitenge, line 48:
```suggestion
        : `allowed_values.values.values`,
```

src/catalogue/category/property/allowedValuesListTextFields.component.tsx
```diff
@@ -0,0 +1,226 @@
...
+  } = useFormContext<
+    AddPropertyMigration | AddCatalogueCategoryWithPlacementIds
+  >();
+
+  const propertyRHF = watch();
+
+  const { fields, append, remove } = useFieldArray({
+    control,
+    name:
+      typeof nestIndex === 'number' && 'properties' in propertyRHF
```
joelvdavies, line 46: I think this could do with a comment.

src/catalogue/category/property/allowedValuesListTextFields.component.tsx
```diff
@@ -0,0 +1,226 @@
...
+              name={
+                typeof nestIndex === 'number' && 'properties' in propertyRHF
+                  ? `properties.${nestIndex}.allowed_values.values.values.${index}`
+                  : `allowed_values.values.values.${index}`
+              }
+              render={({ field: controllerField }) => (
+                <TextField
+                  disabled={allowedValuesIds?.includes(field.av_placement_id)}
+                  id={`list-item-input-${controllerField.value.av_placement_id}`}
+                  label={`List item`}
```
joelvdavies, line 161:
```suggestion
                  label="List item"
```

joelvdavies (commented):
Worked well for me. Noticed a slight bug we may want to fix, allowed number values saves with a value "      ", which isn't a valid number.

src/catalogue/category/property/allowedValuesListTextFields.component.tsx
```diff
@@ -42,6 +42,7 @@ const AllowedValuesListTextFields = (props: {
 
   const { fields, append, remove } = useFieldArray({
     control,
+    // Sets the correct name for the field array based on the type of `nestIndex` and the presence of 'properties' in `propertyRHF`.
```
joelvdavies, line 45: Could you just clarify when this is the case, I can see what the comment says  by reading the line, what isn't so clear is why its nested. I guess just something like `When multiple properties are given allowed values will be nested within those properties` if I am understanding it correctly.

joshuadkitenge: Isn't that implied by the type anyways. The comment would just be describing the parent type form either (AddPropertyMigration | AddCatalogueCategoryWithPlacementIds).

joelvdavies: True, but I didn't see that first and to me that is less obvious than the comment as it is now which is just describing the if statement rather than why the if statement was needed.

joshuadkitenge: I've removed the comment and I don't think there is a need for one as name is determined by the type given to the useForm function in react hook form anyways.

joelvdavies: Sure its clearer once you know what its doing, which is what a comment is supposed to help you understand, but in this case I think it was my fault, it may be clearer to others.

joelvdavies (approved):

PR #1165, opened 2024-12-17
Modify dialogue error messages when editing a catalogue item that has child elements #1136

### LINKED ISSUE #1136: Modify catalogue item edit dialogue error messages when has children
Following on from https://github.com/ral-facilities/inventory-management-system-api/pull/418. We now get error messages like
image
This should be modified on the front end so its more user friendly e.g. manufacturer_id should be replaced by manufacturer.

rowan04:
## Description

Modifies the error message to be more user friendly. Includes displaying the actual manufacturer name.
Adds cypress tests and add/modifies unit tests.
As seen here.
imscatalogueitemediterrormessage

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

Resolves #1136


src/catalogue/items/catalogueItemsDialog.component.tsx
```diff
@@ -386,8 +386,18 @@ function CatalogueItemsDialog(props: CatalogueItemsDialogProps) {
...
+                  // find the name of the manufacturer, so it can be used in the error message
+                  const manufacturerName = manufacturerList?.find(
+                    (manufacturer) => manufacturer.id === selectedCatalogueItem.manufacturer_id
+                  ) || null
+                  // add the manufacturer name into the error message
+                  const message = response.detail.replace(
+                    "so the following fields cannot be updated: manufacturer_id, properties",
+                    ("so you cannot update the properties, and the manufacturer cannot be changed from "
+                      + manufacturerName?.name)
+                  )
```
joshuadkitenge, line 398: Instead of replacing the text directly in the backend response, consider moving this string transformation entirely to the frontend. This approach will ensure consistency in the response formatting, prevent future inconsistencies, and make it easier to manage localization or text changes.

src/catalogue/items/catalogueItemsDialog.component.test.tsx
```diff
@@ -1084,7 +1084,39 @@ describe('Catalogue Items Dialog', () => {
...
+        manufacturer_id: '3',
+      });
+
+      await waitFor(() => {
+        expect(
+          screen.getByText(
+            'Catalogue item has child elements, so you cannot update the properties, '
+              + 'and the manufacturer cannot be changed'
+          )
+        ).toBeInTheDocument();
```
joshuadkitenge, line 1115: The error message is missing the name of the manufacturer

cypress/e2e/with_mock_data/catalogueItems.cy.ts
```diff
@@ -581,15 +581,32 @@ describe('Catalogue Items', () => {
...
+    cy.findByLabelText('Manufacturer *').type('Man{downArrow}{enter}');
+    cy.findByRole('button', { name: 'Next' }).click();
+
+    cy.findByRole('button', { name: 'Finish'}).click();
+    cy.findByRole('dialog')
+      .should('be.visible')
+      .within(() => {
+        cy.contains('Catalogue item has child elements, so you cannot update the properties, '
+          + 'and the manufacturer cannot be changed')
+      });
```
joshuadkitenge, line 594: same here

joshuadkitenge (changes requested):

asuresh-code (changes requested):
In the case of catalogue items, I think the only child elements are items themselves? If that's the case I think it's more user friendly to say it has items, rather than child elements.

rowan04:
> In the case of catalogue items, I think the only child elements are items themselves? If that's the case I think it's more user friendly to say it has items, rather than child elements.

I have changed it from `child elements` to `items` for the actual usage, but I am unable to do this for the tests as they all use the phrase `child elements` in their queries.

asuresh-code:
A very minor suggestion, but I think the error message would be better if you put the error "cannot update" before the reasoning "catalogue item has items", so something like this.

`"Unable to update properties and manufacturer (Digital Dynamics Inc.) as the catalogue item contains associated items."`

I think in most error messages, it says the failure first, and then the reasoning.

asuresh-code:
Actually, I've just tested it again without the api (so just using mock data), and the error doesn't seem to be appearing. I'm not sure if this is meant to be the case or not?

src/catalogue/items/catalogueItemsDialog.component.tsx
```diff
@@ -526,6 +526,24 @@ function CatalogueItemsDialog(props: CatalogueItemsDialogProps) {
...
+    if (message?.includes('child elements')) {
+      // find the name of the manufacturer, so it can be used in the error message
+      const manufacturerName = manufacturerList?.find(
+        (manufacturer) => manufacturer.id === selectedCatalogueItem?.manufacturer_id
+      ) || null;
+      // add the manufacturer name into the error message
+      const childElementsMessage = message.replace(
+        "child elements, so the following fields cannot be updated: manufacturer_id, properties",
+        ("items, so you cannot update the properties, and the manufacturer cannot be changed from "
+          + manufacturerName?.name)
```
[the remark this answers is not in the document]
rowan04: yes that's definitely clearer! It's because i was originally only replacing the second half of the string, but yeah now its replacing everything except the first 3 words, so this makes more sense

rowan04:
> Actually, I've just tested it again without the api (so just using mock data), and the error doesn't seem to be appearing. I'm not sure if this is meant to be the case or not?

Same, I'm not sure why this is the case.

rowan04:
> A very minor suggestion, but I think the error message would be better if you put the error "cannot update" before the reasoning "catalogue item has items", so something like this.
> 
> `"Unable to update properties and manufacturer (Digital Dynamics Inc.) as the catalogue item contains associated items."`
> 
> I think in most error messages, it says the failure first, and then the reasoning.

done

asuresh-code (approved):
LGTM, I checked `develop`, and the error doesn't appear there either, so I assume it's not an issue, unless @joshuadkitenge thinks otherwise.

src/catalogue/items/catalogueItemsDialog.component.tsx
```diff
@@ -526,6 +526,21 @@ function CatalogueItemsDialog(props: CatalogueItemsDialogProps) {
...
+      ) || null;
+      // add the manufacturer name into the error message
+      const childElementsMessage = "Unable to update catalogue item properties and manufacturer ("
+        + manufacturerName?.name + "), as the catalogue item has associated items.";
+      return (childElementsMessage);
+    } else {
+      return (message);
+    };
+  };
+
```
joshuadkitenge, line 543: The error message for child elements should be set on line 390 in the `setErrorPropertiesStep ` not on line 1020. You shouldn't need to use a function either.

joshuadkitenge (changes requested):

rowan04:
image

rowan04:
the failing tests are all unrelated to my changes

joshuadkitenge:
> the failing tests are all unrelated to my changes
you just need to rerun the test. I'm going to fix the flakiness in a separate pr

PR #1228, opened 2025-01-08
Handle 404 errors with React Router for manufacturers

joshuadkitenge:
## Description

- Create loader for manufacturer and the error pages 
- The createBrowerRouter need to be created after msw is build for dev 

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Note

The layout is only applied to the child routes. If the error page is done further done the sub tree it would trigger unnecessary API calls in the Layout component (getCatalogueItem etc ). This is the reason why it done at the top level.


src/App.tsx
```diff
@@ -92,25 +100,39 @@ const router = createBrowserRouter([
       {
         path: paths.manufacturers,
         Component: ManufacturerLayout,
+        loader: manufacturerLayoutLoader(queryClient),
+        ErrorBoundary: ManufacturerLayoutErrorComponent,
```
joelvdavies, line 104: Not sure how react router handles this, but couldn't this be an `errorElement: ManufacturerErrorComponent` (or even just the error page directly) so that it uses existing layout when rendering it avoiding the need to also specifying the BaseLayoutHeader inside the error component as well.

src/manufacturer/manufacturerLayout.component.test.tsx
```diff
@@ -60,3 +65,70 @@ describe('Manufacturer Layout', () => {
...
+    await user.click(homeButton);
+
+    expect(mockedUseNavigate).toHaveBeenCalledTimes(1);
+    expect(mockedUseNavigate).toHaveBeenCalledWith('/manufacturers');
+
+    expect(view.asFragment()).toMatchSnapshot();
+  });
+});
+
+describe('manufacturerLayoutLoader', () => {
```
joelvdavies, line 109: This name is inconsistent with the describe above.

joelvdavies (commented):
Worked well, just some suggestions.

src/api/manufacturers.tsx
```diff
@@ -75,15 +76,18 @@ const getManufacturer = async (id: string): Promise<Manufacturer> => {
     .get(`/v1/manufacturers/${id}`)
     .then((response) => response.data);
 };
-
-export const useGetManufacturer = (
-  id?: string | null
-): UseQueryResult<Manufacturer, AxiosError> => {
-  return useQuery({
+export const getManufacturerQuery = (id?: string | null, loader?: boolean) =>
```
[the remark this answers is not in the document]
joshuadkitenge: getManufacturer is the axios hook name above. so it need to be getManufacturerQuery (matches react query example).
```
const getManufacturer = async (id: string): Promise<Manufacturer> => {
  return imsApi
    .get(`/v1/manufacturers/${id}`)
    .then((response) => response.data);
};
```

src/manufacturer/manufacturerLayout.component.test.tsx
joshuadkitenge: This is on purpose as this is a function and there other ones are components

src/App.tsx
joshuadkitenge: It doesn't work like that. The layout is only applied to the child routes.

joelvdavies: Ah yeah I see it would only have been the case if BaseLayout was at the root of the rest of the routes instead of inside of them.

src/api/manufacturers.tsx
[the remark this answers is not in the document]
joelvdavies: 🤦‍♂️ I didn't see the line above, thanks for the clarification.

joelvdavies (approved):

PR #1284, opened 2025-02-04
Add body and meta types for upload Dialogs

joshuadkitenge:
## Description

Currently the types are just using Meta and Body from Uppy. These should be specific for attachments and images 

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

### DIFF
--- a/src/api/api.types.tsx
+++ b/src/api/api.types.tsx
@@ -218,16 +218,25 @@ export interface Item
   usage_status: string;
   properties: Property[];
 }
+// ------------------------------------ OBJECT STORAGE API -----------------------------------------
 
-// ------------------------------------ ATTACHMENTS ------------------------------------------------
-
-// This is AttachmentPost on the object-store-api
-export interface AttachmentPostMetadata {
+export interface ObjectFileUploadMetadata {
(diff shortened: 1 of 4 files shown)


src/api/api.types.tsx
```diff
@@ -218,17 +220,22 @@ export interface Item
...
-// This is AttachmentPost on the object-store-api
-export interface AttachmentPostMetadata {
+export interface ObjectFileUploadMetadata {
   entity_id: string;
   file_name: string;
   title?: string | null;
   description?: string | null;
 }
 
+export interface UppyUploadMetadata extends ObjectFileUploadMetadata, Meta {}
```
joelvdavies, line 232: I think this and the one below are best placed in the app.types - I don't think they api related at this point or perhaps shouldn't mention uppy. Although the Meta and Body types are just basic dictionaries by the looks of it (Record<string, unknown>), so not sure if they need to be extended here as its basically just saying it can have anything else in addition to what is specified which I don't think is true?

joelvdavies (approved):
Thanks for the clarification, LGTM

PR #1295, opened 2025-02-07
Investigate flaky ci tests for upload attachments dialog

asuresh-code:
## Description

A temporary PR to investigate issues with the flaky test in `UploadAttachmentsDialog.component.tsx`. I have not been able to recreate the same errors in my VM, so I'm investigating on the Github CI.

### DIFF
--- a/src/common/attachments/uploadAttachmentsDialog.component.test.tsx
+++ b/src/common/attachments/uploadAttachmentsDialog.component.test.tsx
@@ -96,26 +96,28 @@ describe('Upload attachment dialog', () => {
     // Checks if file extension is displayed. If it's editable, actual value will disappear after editing.
     expect(await screen.findByText('.txt')).toBeInTheDocument();
 
-    const title = screen.getByRole('textbox', { name: 'Title' });
-    const description = screen.getByRole('textbox', { name: 'Description' });
+    const description: HTMLInputElement = screen.getByRole('textbox', {
+      name: 'Description',
+    });
+    const title: HTMLInputElement = screen.getByRole('textbox', {
+      name: 'Title',
+    });
 
     expect(await screen.findByDisplayValue('test1')).toBeInTheDocument();
 
-    await user.type(title, 'test title');
+    fireEvent.click(title);
 
-    await user.clear(description);
-    await user.click(description);
-    await user.paste('test description');
+    fireEvent.change(title, {
+      target: { value: 'test title' },
+    });
 
-    await user.click(title);
-    await user.click(description);
-    await user.tab();
+    fireEvent.click(description);
+
+    fireEvent.change(description, {
+      target: { value: 'test description' },
+    });
 
     expect(await screen.findByText('.txt')).toBeInTheDocument();
-
-    expect(
-      await screen.findByDisplayValue('test description')
-    ).toBeInTheDocument();
 
     await user.click(await screen.findByText('Save changes'));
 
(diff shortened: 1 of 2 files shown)


src/common/attachments/uploadAttachmentsDialog.component.test.tsx
```diff
@@ -100,14 +100,19 @@ describe('Upload attachment dialog', () => {
...
-    expect(await screen.findByDisplayValue('test title')).toBeInTheDocument();
+    await user.clear(description);
+    await user.click(description);
+    await user.paste('test description');
+
+    await user.click(title);
     await user.click(description);
+    await user.tab();
+
+    expect(await screen.findByText('.txt')).toBeInTheDocument();
```
joshdimanteto, line 114: did you try increasing the timeout or use fireEvent instead (in some tests we use fireEvent as it is less flaky ) ?

src/common/attachments/uploadAttachmentsDialog.component.tsx
```diff
@@ -28,6 +28,7 @@ const UploadAttachmentsDialog = (props: UploadAttachmentsDialogProps) => {
   const [fileMetadataMap, setFileMetadataMap] = React.useState<
     Record<string, string>
   >({});
+  console.log(`INITIAL FileMetadataMap: ${JSON.stringify(fileMetadataMap)}`);
```
joshdimanteto, line 31: Remove the console.log statements

joshdimanteto (changes requested):

src/common/attachments/uploadAttachmentsDialog.component.test.tsx
asuresh-code: I did not try using fireEvent. I am pretty sure I tried increasing the timeout although I don't remember exactly if I did.

asuresh-code: I've tried using fireEvent (see latest commit), but wasn't able to get it to work with uppy. I tried using it without `fireEvent.click()` and by using `screen.getByLabelText` instead, and by placing the `screen.get...` portion of code in the fireEvent instead of storing the value in a variable.

asuresh-code: So what I've gathered so far when looking into is:
- When using `fireEvent`, the value in the input field does update, but the value in the request stays blank `undefined`.
- If I add a dispatch input event to both fields after using `fireEvent.change/focus`, then the 1st field is correctly set in the request `"test title/description"`, but the second field is still undefined.
- Adding a timeout to checks like the one below, does not fix it. These checks pass fine as the input field has the correct value, but the request will still have undefined for both or the last field updated.
```
await waitFor(() => {
      expect(description).toHaveValue('test description');
    });
```

asuresh-code: In my latest commit, I've created a new userEvent with a delay in the typing, which seems to have fixed the issue (0 failures in 16 test runs locally).

As a result, I've been able to remove the excess user interactions.

src/common/attachments/uploadAttachmentsDialog.component.test.tsx
```diff
@@ -96,18 +96,24 @@ describe('Upload attachment dialog', () => {
...
-    await user.type(title, 'test title');
+    const delayedUser = userEvent.setup({ delay: 20 });
 
-    expect(await screen.findByDisplayValue('test title')).toBeInTheDocument();
-    await user.click(description);
+    await delayedUser.type(title, 'test title');
+
+    await delayedUser.type(description, 'test description');
+
+    expect(await screen.findByText('.txt')).toBeInTheDocument();
```
joshdimanteto, line 115: Can you add a comment explaining what you think is happening and why you chose the fix. Also make sure to rerun the unit test a couple of times on the CI to make sure that the implementation has worked

joshdimanteto (changes requested):

joshdimanteto (approved):

PR #1297, opened 2025-02-17
Feature/remove primary images dialog #1096

asuresh-code:
## Description

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}


src/common/images/primaryImage.component.tsx
```diff
@@ -70,8 +92,37 @@ export interface PrimaryImageProps {
...
+        .then((data: APIImageWithURL) => {
+          setImageData(data);
+        })
+        .catch((error: AxiosError) => {
+          handleIMS_APIError(error);
+        });
+    } else {
+      setImageData(undefined);
+    }
+  }, [primaryImageExists, imagesData, setImageData]);
```
joshdimanteto, line 121: Could you explain why queryClient.fetchQuery was used in this file? I can see why it makes sense in the download function—do you know the reasoning behind that as well?

joshdimanteto (changes requested):
This needs to be based on viewing primary images. It can't be properly tested without the view primary images in the frontend.

src/common/images/primaryImage.component.tsx
asuresh-code: At the time, I wanted to conditionally call `getImage`, only if a primary image existed. I didn't want to call it every time because passing an empty string if the primary image did not exist, would form a valid request to get a list of all images. You could probably parse the result to see if the primary image exists or not after (i.e if it returns a list or single image), but that seemed a bit flaky.

So to conditionally call `getImage`, I had the option to either add an enabled parameter to `useGetImage`, or to add it to `getImage` and use the `queryClient` in a conditional statement. I opted for the second option, because I think the same question of how to implement this in the download issue came up, and you suggested to use the `queryClient`, over adding an enabled parameter to the `useApiFunction`.

Now, technically the `queryClient getImage` is not actually required for this remove primary image PR; the download url is not needed to remove the primary image. I added this in because at the time it was required for the view primary Image PR, so I though it would require less refactoring to just add this in in this PR (The view Primary Images PR copies this implementation). However, I think you mentioned earlier that we may just use the thumbnail for viewing the primary image in the landing page (not the lightbox). If that's the case then this additional call could actually just be removed, and we just use the result from the 1st get images call in this file.

I don't remember exactly why the decision was made for the download function; I think it was because we also wanted to only get the image conditionally (In this case when the user clicked the download button, since we didn't want to needlessly download it if the user closed the dialog). I think I originally added enabled to the `useGetImage` function to conditionally call, but you suggested to use queryClient instead, since it meant I did not need to use useEffect, making it less flaky.

joshdimanteto (approved):
LGTM

PR #1302, opened 2025-03-10
Improve the card views

joshdimanteto:
## Description

- Improve the resizing of the card views
- Increase the size of the tables that are on landing pages
- Allow for attachment and images to be fullscreen 

## Testing instructions

- [ ] Review code
- [ ] Check Actions build


src/catalogue/category/catalogueCardView.component.tsx
```diff
@@ -516,49 +509,64 @@ function CatalogueCardView() {
...
+              sx={{
+                justifyContent: 'left',
+                paddingLeft: 0.5,
+                position: 'sticky',
+                top: 0,
+                backgroundColor: 'background.default',
+                zIndex: 1000,
+                width: '100%',
+                paddingTop: 2.5,
+                paddingBottom: 2.5,
```
asuresh-code, line 536: For the value `position: 'sticky'`, I think it could be improved if it was conditional based on the viewport width, so that if the window is narrow, it is no longer sticky. This is because if the window is narrow, the filters take up the whole height of the window, and you can't see the content.

Admittedly, I doubt anyone would have that narrow a window, but I don't think it would be a hard fix to implement
image

src/catalogue/category/catalogueCardView.component.tsx
```diff
@@ -516,49 +509,64 @@ function CatalogueCardView() {
...
-            item
-            overflow={'auto'}
-          >
-            <Grid item width={'100%'}>
-              <MRT_TopToolbar table={table} />
-            </Grid>
-            <Grid item container direction="column" alignItems="center">
+          <Grid item width="100%" sx={{ flexShrink: 0 }}>
+            <MRT_TopToolbar table={table} />
+          </Grid>
```
asuresh-code, line 518: What is explicitly setting `flexShrink` to `0` doing that it was not before?

I can see what it's meant to do in this demo, however when comparing develop to this branch, I couldn't find any difference in the way the toolbar resizes.

src/catalogue/category/catalogueCardView.component.tsx
```diff
@@ -516,49 +509,64 @@ function CatalogueCardView() {
...
+              direction="column"
+              alignItems="center"
+              display={!isCollapsed ? 'none' : undefined}
+              sx={{
+                justifyContent: 'left',
+                paddingLeft: 0.5,
+                position: 'sticky',
+                top: 0,
+                backgroundColor: 'background.default',
+                zIndex: 1000,
```
asuresh-code, line 533: Why is a `zIndex` of `1000` used? Again, it wasn't clear to me how it was different to the develop branch without it.

asuresh-code (changes requested):

src/catalogue/category/catalogueCardView.component.tsx
joshdimanteto: This is not an issue as they can just close the filters

src/catalogue/category/catalogueCardView.component.tsx
joshdimanteto: This prevents the grid from shrinking (https://developer.mozilla.org/en-US/docs/Web/CSS/flex-shrink). There is not meant to be any visible changes. The way a did it before was an OK methodology the main issue with it is that it only works in one view as i used fixed heights. This methodology reduces that.

src/catalogue/category/catalogueCardView.component.tsx
joshdimanteto: ## DEV
image

## Branch
image


Here is the docs on zIndex (https://developer.mozilla.org/en-US/docs/Web/CSS/z-index). This is to ensure the the filters are always visible

asuresh-code (changes requested):
One thing I've just noticed now after switching to the mock data, is that the initial catalogue card views page is extra wide for some reason. This only occurs on the home page, and it does not occur when using the api with and without scigateway enabled.

out of bounds cards

joshdimanteto:
> One thing I've just noticed now after switching to the mock data, is that the initial catalogue card views page is extra wide for some reason. This only occurs on the home page, and it does not occur when using the api with and without scigateway enabled.
> 
> out of bounds cards [ out of bounds cards ](https://private-user-images.githubusercontent.com/69875854/421355931-7d78eb4f-de10-4cf5-8c95-9bd655a0ae5b.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NDE2OTU5OTYsIm5iZiI6MTc0MTY5NTY5NiwicGF0aCI6Ii82OTg3NTg1NC80MjEzNTU5MzEtN2Q3OGViNGYtZGUxMC00Y2Y1LThjOTUtOWJkNjU1YTBhZTViLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNTAzMTElMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjUwMzExVDEyMjEzNlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWY5NjllYjVjODg0OTMwZmM2NDE2YTQzYzI1OTFlMTkxNmYxYjdjYjk5NDM3ZmQzZGZkZjgwYzRkN2Y1M2U1MDQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.i5JdTBenvxLLEVlYrtyq5C-Lp9ChlvQtMEHJamkdzXc) [ ](https://private-user-images.githubusercontent.com/69875854/421355931-7d78eb4f-de10-4cf5-8c95-9bd655a0ae5b.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NDE2OTU5OTYsIm5iZiI6MTc0MTY5NTY5NiwicGF0aCI6Ii82OTg3NTg1NC80MjEzNTU5MzEtN2Q3OGViNGYtZGUxMC00Y2Y1LThjOTUtOWJkNjU1YTBhZTViLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNTAzMTElMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjUwMzExVDEyMjEzNlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWY5NjllYjVjODg0OTMwZmM2NDE2YTQzYzI1OTFlMTkxNmYxYjdjYjk5NDM3ZmQzZGZkZjgwYzRkN2Y1M2U1MDQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.i5JdTBenvxLLEVlYrtyq5C-Lp9ChlvQtMEHJamkdzXc)
> 
> Edit: This issue seems to be introduced in your second commit, [c0ec6a4](https://github.com/ral-facilities/inventory-management-system/commit/c0ec6a4478156ae54969faf652aa32eca586e271)

Good catch 👍

PR #1304, opened 2025-03-10
Store lightbox state in url

joshdimanteto:
## Description
- Store the lightbox state in the url, by using the current id of the imageGallery
- remove unnecessary condition for delayedLoader in lightbox

## Testing instructions

- [ ] Review code
- [ ] Check Actions build


asuresh-code (changes requested):
Looks good and works well, just have a few comments to make:
- I think changing the url param variable from `image` to something like `imageId` makes it more clear what it is, and reduces the chance of any conflict.
- Whilst there is full code coverage, I think adding a test that opens an image in the lightbox and checks that the url has correctly updated would be a good idea.

joshdimanteto:
> Looks good and works well, just have a few comments to make:
> 
>     * I think changing the url param variable from `image` to something like `imageId` makes it more clear what it is, and reduces the chance of any conflict.
> 
>     * Whilst there is full code coverage, I think adding a test that opens an image in the lightbox and checks that the url has correctly updated would be a good idea.

Thanks for the feedback! Here are my thoughts on both points:

1. **Changing `image` to `imageId`**  
   - `image` works well in this context and is unlikely to cause any issues. If a conflict does arise in the future, we can address it then. For now, keeping it as `image` keeps things simple.  

2. **Adding a test for lightbox URL update**  
   - The key concern is whether the lightbox opens and functions correctly, which is already covered. Since the lightbox being interactive and usable inherently confirms that the URL updates as expected, adding a separate test for this wouldn’t add much value.

asuresh-code (approved):
LGTM

PR #1325, opened 2025-03-21
Move expected lifetime into items #1323

### LINKED ISSUE #1323: Move expected lifetime from catalogue items to items
CLF requested that it be moved to items but remain user entered. This will cover the fact it really changes per catalogue item-system pairing by forcing users to add it for every single item instead until calculating the lifetime from the history is implemented.

See corresponding backend issue https://github.com/ral-facilities/inventory-management-system-api/issues/479.

joelvdavies:
## Description

Corresponding PR for backend one https://github.com/ral-facilities/inventory-management-system-api/pull/481. 

Feel free to reassign to fix the remaining tests/address any changes. I have done as much as I could before going on leave. Otherwise the work would also need to be completely removed as well when they could at least start populating it themselves in v2.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

Closes #1323


joelvdavies:
Some of the remaining test failures could be due to scrolling from some of the tables missing/gaining a column. If it helps I had to change some scrolling commands by 200px for them to function in other places.

joelvdavies:
One e2e api test is failing, not sure why. Also could potentially modify a test to cover last bit of coverage missing, but it minor and didn't get time.

joshdimanteto (commented):
Works well, I'm going to create another pr that add some extra features once that is ready this can be merged in together.

- Allow user to change the expected life when moving an item, also investigate if they my need other fields 
    - _There is no good way of doing this without making the dialog more cumbersome_ 
    - _I'll do the second option which is to add the edit item dialog to item rows on the systems items table_   

- Do some mean calculation in the aggregate cells (internal MRT logic )

joshdimanteto:
> One e2e api test is failing, not sure why. Also could potentially modify a test to cover last bit of coverage missing, but it minor and didn't get time, otherwise should be ready for review.

That test should pass once the backend changes have been merged in. The expected lifetime value doesn't exist in items collection on ims-api dev so the duplicate (for items) test fails

joelvdavies:
Just tried it and it passed here https://github.com/ral-facilities/inventory-management-system/actions/runs/14351716865/job/40231806003?pr=1325 (just an unrelated flaky unit test)

joelvdavies:
This is now no longer needed, after discussion the requested change has been reversed again.

PR #1334, opened 2025-04-03
Configuration of Backend Errors in Uppy

### LINKED ISSUE #1275: Configure Maximum file limit for an given entity (images)
#### Tasks:
- [ ] Configure the maximum  upload file limit in the **settings files**.
- [ ] Ensure this configuration is implemented in the **Docker files** for proper deployment.

### LINKED ISSUE #1276: Configure Maximum Upload file limit for Attachments
#### Tasks:
- [ ] Configure the maximum  upload file limit in the **settings files**.
- [ ] Ensure this configuration is implemented in the **Docker files** for proper deployment.

asuresh-code:
## Description

Sets Attachments and Images maximum upload file limit error messages.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #1275 
closes #1276


src/api/api.tsx
```diff
@@ -126,6 +126,12 @@ export function uppyOnAfterResponse(xhr: XMLHttpRequest) {
           else resolve();
         });
       });
+    } else if (retryCount === 3 || (xhr.status >= 400 && xhr.status < 600)) {
+      let errorMessage = (JSON.parse(xhr.responseText) as APIError).detail;
+      if (errorMessage.includes('Limit for the maximum number of images')) {
```
rowan04, line 131: i wonder if it could do with a comment here to explain why attachments aren't handled in this file?

rowan04 (commented):
Code looks good to me, but i didn't test functionality. Could probably do with some tests too?

src/utils.tsx
```diff
@@ -551,3 +551,17 @@ export function downloadFileByLink(url: string, filename: string): void {
...
+    returnMessage = 'Maximum number of files reached.';
+  } else if (errorMessage.includes('does not contain the correct extension')) {
+    returnMessage = 'File extension does not match content type.';
+  } else if (errorMessage.includes('is not supported')) {
+    returnMessage = 'Content type not supported.';
+  } else if (errorMessage.includes('not a valid image')) {
+    returnMessage = 'File given is not a valid image.';
+  }
+  return returnMessage;
+}
```
joshdimanteto, line 567: Can all of these errors actually be triggered from the frontend? It might be worth confirming that these specific error messages are returned in scenarios the user can encounter on the client side. Otherwise, we might be handling edge cases that never surface in the UI.

asuresh-code:
I'm currently investigating an issue where the correct error messages appear when I use `docker compose up` to run the front-end, but they errors do not load when I use `yarn dev`. I have tried deleting the node modules and reinstalling with yarn install.

chech
Successful error using docker compose up

c
Unsuccessful error when using yarn dev

joshdimanteto:
run `docker compose build` and it should work

asuresh-code:
> run `docker compose build` and it should work

I tried running this, and it didn't really change the outcomes; it still works using `docker compose up`, and fails using `yarn dev`

src/utils.tsx
asuresh-code: Good point, below I've put a small analysis on whether the errors could be triggered on the Front-End. To summarise, most of the errors are triggerable on some Operating Systems, but not others due to the differences in how they derive the MIME type. The Max file limit error is the only one that is triggerable on all OSes, and the unexpected errors are the only ones that should not occur on any OS.

**Always Triggerable:**
- **Max file limit:** Simply uploading an image/attachment after the limit for an entity has been reached.

**Triggerable Depending On Operating System:**
- **File Extension does not match content type:** This should be impossible for users using a **Windows** device as the MIME type is derived from the file extension (so a mismatch should be impossible). However, on **Linux** MIME types come from inspecting the file contents, and on **MacOS** it's derived from a mixture of file metadata, UTI (Uniform Type Identifier), and file contents, so this can be triggered on the Front End on these devices.

- **Content Type not supported:** On the front-end using Uppy, we check if files are a supported type **based on their MIME type**, but on the back-end we raise the error **based on its file extension**. This leads to a similar situation as the error above. On **Windows** it should be impossible (the content type and file extension are tightly coupled, so any file that passes the uppy filter will not raise an error on the back end). However again, for **Linux** and **MacOS** this error could be raised. A user could change the file extension of a `jpeg` to `image.gif`. It would pass the uppy filter as the MIME type would be unaffected, but it would fail on the back-end as the `.gif` file extension is unsupported.

- **Invalid Image:** This is triggerable on **Windows** (you could change `movie.mp4` to `movie.jpeg`, which changes the file extension + MIME type). It shouldn't be triggerable on **Linux**? (I believe Linux inspects the file contents for the MIME type, so a non-image wouldn't make it past the uppy filters). It might be possible on **MacOS** since it uses a mixture of file contents/UTI/and other metadata.

**Should be "Impossible" to Trigger**
- **Unexpected Error:** This covered 2 errors the back-end raised, the `MissingRecordError` from getting an image after uploading it, and `InvalidObjectIdError` you get if the entity ID for the image/attachment that is stored is not valid. Neither of these should be able to be occur both on the front-end and if you are just using the endpoints, so I added this message to catch these 2 scenarios instead of using Uppy's network error message.

asuresh-code: They should be, except for the Invalid Image Error. 

There is a PR on the back-end to add a restriction for supported attachment file types.
 
There currently isn't a check for the file extension not matching the content type when creating attachments, which I think may be a mistake on the back end.

joshdimanteto: All these cases need to be tested (unit and mocked e2e test)

joshdimanteto:
My suspicion is that if the Docker container is running at the same time, the promises might get caught in the wrong place. Ideally, you should always shut down the Docker container (docker compose down) when switching between Docker and yarn dev.

if this error still persist run:

```
docker compose down
yarn cache clean
rm -rf node_modules/
yarn install
yarn dev
```

joshdimanteto (approved):

PR #1335, opened 2025-04-03
fix: resolve close button issue in upload attachments dialog

joshdimanteto:
## Description
https://github.com/ral-facilities/inventory-management-system/pull/1334#issuecomment-2774838571

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

### DIFF
--- a/src/common/uppy.utils.tsx
+++ b/src/common/uppy.utils.tsx
@@ -150,6 +150,7 @@ export function useMetaFields<M extends Meta, B extends Body>(): MetaFields<
   return metaFieldsData;
 }
 
+// https://github.com/transloadit/uppy/blob/229902eb17cbb9f866921cf232d94b25eb584efd/packages/%40uppy/status-bar/src/StatusBar.tsx#L24
 export function getUploadingState(
   error: unknown,
   isAllComplete: boolean,
--- a/src/common/uppy.utils.tsx
+++ b/src/common/uppy.utils.tsx
@@ -1,6 +1,8 @@
 import { Box, styled } from '@mui/material';
 import { Body, Meta, type UppyFile } from '@uppy/core';
 import { DashboardState } from '@uppy/dashboard/lib/Dashboard';
+import statusBarStates from '@uppy/status-bar/lib/StatusBarStates';
+import { type StatusBarUIProps } from '@uppy/status-bar/lib/StatusBarUI';
 import type { VNode } from 'preact';
 import React from 'react';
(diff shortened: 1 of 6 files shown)


src/common/uppy.utils.tsx
```diff
@@ -148,16 +150,47 @@ export function useMetaFields<M extends Meta, B extends Body>(): MetaFields<
...
+  isAllComplete: boolean,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  recoveredState: any,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  files: Record<string, UppyFile<any, any>>
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+): StatusBarUIProps<any, any>['uploadState'] {
+  if (error) {
+    return statusBarStates.STATE_ERROR;
+  }
```
joshdimanteto, line 164:
```suggestion
// https://github.com/transloadit/uppy/blob/229902eb17cbb9f866921cf232d94b25eb584efd/packages/%40uppy/status-bar/src/StatusBar.tsx#L24
export function getUploadingState(
  error: unknown,
  isAllComplete: boolean,
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  recoveredState: any,
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  files: Record<string, UppyFile<any, any>>
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
): StatusBarUIProps<any, any>['uploadState'] {
  if (error) {
    return statusBarStates.STATE_ERROR;
  }
```

asuresh-code (commented):
Works well, my only comment is if you think there should be some tests added?

I can see that the changes in uppy.utils are pretty much just a copy of the internal uppy implementation, so it may not be needed there. However, maybe there should be a test in both of the dialog components, specifically for returning (preventing dialog close) when a file is being uploaded.

joshdimanteto:
> Works well, my only comment is if you think there should be some tests added?
> 
> I can see that the changes in uppy.utils are pretty much just a copy of the internal uppy implementation, so it may not be needed there. However, maybe there should be a test in both of the dialog components, specifically for returning (preventing the dialog closing) when a file is being uploaded.

Yeah, it is exactly that as this is a function copied from uppy  doesn't need any testing

joshdimanteto:
I'll look into the testing later. This need to be in before the demo

PR #1339, opened 2025-04-07
Setup Object-Storage-API on CI

### LINKED ISSUE #1088: Set up object-storage-api on CI
- using the docker image object-storage-api

asuresh-code:
## Description

Sets up the object storage API in the CI to implement e2e api tests for images and attachments.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #1088


.github/workflows/ci-build.yml
```diff
@@ -77,7 +77,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Clone api repo
+      - name: Clone IMS api repo
         uses: actions/checkout@cbb722410c2e876e24abbe8de2cc27693e501dcb
         with:
           repository: ral-facilities/inventory-management-system-api
```
joelvdavies, line 83: I notice that you used `path: object-storage-api` below, if you use similar here, can you avoid the `Move IMS api repo` step below? Or maybe it would have to be `../inventory-management-system-api` to have the effect? Presumably it would replace
```
          cd ..
          mkdir inventory-management-system-api
          mv -v inventory-management-system/* inventory-management-system-api/
```
And the remaining cd could go under a different step or its own one.

joelvdavies (commented):
One minor suggestion and one I thought might simplify things but am not sure if it would work. Looks like it works well though.

.github/workflows/ci-build.yml
asuresh-code: I don't think I can remove the move step for either APIs. I tried it now and got this error on the CI during the clone job.

`Repository path '/home/runner/work/inventory-management-system/inventory-management-system-api' is not under '/home/runner/work/inventory-management-system/inventory-management-system'`

It does mention in the comment above Moving the IMS api-repo that is it is necessary, and that the clone step cannot clone into a different directory. 

I don't think the `path: object-storage-api` was necessary and I have now removed it.

joelvdavies: Sure thanks for trying, I thought it might be the case, but it was a while ago when I wrote that comment so wasn't sure if it was still needed, but seems the path was doing something different to what I thought 😄

joelvdavies (dismissed):
LGTM, Just a note the PR description could probably do with updating, especially to close https://github.com/ral-facilities/inventory-management-system/issues/1088. 

@joshdimanteto Is https://github.com/ral-facilities/inventory-management-system/issues/1287 a duplicate?

joshdimanteto:
Yeah, I've closed the pr.

joshdimanteto:
LGTM, hold off merging this pr until after the demo

.github/workflows/ci-build.yml
joelvdavies, line 1: Just noticed, could you also modify the name of  `Output docker logs (api)` as well to match, and add one for the object-storage-api as well, it should be useful for debugging.

joshdimanteto (dismissed):

.github/workflows/ci-build.yml
asuresh-code: This suggestion has now been implemented in the e2e-api attachments tests PR

joshdimanteto (approved):

joelvdavies (approved):

PR #1342, opened 2025-04-14
fix: set environment variables correctly in Docker entrypoint

joshdimanteto:
## Description

Updated the Docker entrypoint script to ensure environment variables are set properly. This resolves issues with incorrect configurations during container startup.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Check that this set the variables correctly


docker/docker-entrypoint.sh
```diff
@@ -4,14 +4,22 @@
...
+  --argjson maxImageSizeBytes $MAX_IMAGE_SIZE_BYTES \
+  --arg pluginHost "$PLUGIN_HOST" \
+  '.imsApiUrl = $imsApiUrl |
+   .osApiUrl = $osApiUrl |
+   .maxAttachmentSizeBytes = $maxAttachmentSizeBytes |
+   .attachmentAllowedFileExtensions = $attachmentAllowedFileExtensions |
+   .imageAllowedFileExtensions = $imageAllowedFileExtensions |
+   .maxImageSizeBytes = $maxImageSizeBytes |
+   .pluginHost = $pluginHost' \
+  public/inventory-management-system-settings.example.json > config.json
```
joelvdavies, line 22:
```suggestion
  public/inventory-management-system-settings.example.json > "$TEMPFILE"
```

joshdimanteto: the path was also incorrect so i changed that in the lastest commit

joelvdavies (approved):
Tested and works well with docker-ims.

PR #1344, opened 2025-04-14
fix: prevent image deletion when file is removed from upload dialog

joshdimanteto:
## Description
Ensure that removing a file from the upload dialog does not unintentionally delete successfully upload images. This resolves issues with accidental image loss during file upload.

Do not broadcast the delete error as this is a silent error 

https://github.com/user-attachments/assets/69065f1e-3ea8-4e84-a336-04063a0f701b

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Look at video of current dev


joshdimanteto:
The error handling currently uses setState to delete the metadata. This might be unnecessary if we can rely on Uppy’s internal state instead.

- [ ] Investigate whether the fallback delete metadata functionality can be done using Uppy’s internal state

PR #1417, opened 2025-05-08
Adding conditional github actions job execution

asuresh-code:
## Description

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

connect to {issue number}


LICENSE
joelvdavies, line 1: I think there are some unintended changes here, in the README and in some of the tests below.

asuresh-code: Thanks for pointing it out; I've undone the extra changes in random files (they were for testing the job execution), and I've put 3 commits for the different setup configurations.

joelvdavies (changes requested):
Currently testing option 3, just noticed that a change to a unit test file itself doesn't currently get picked up as a change.

PR #1424, opened 2025-05-22
Upgrade MUI to v6 #915

### LINKED ISSUE #915: Update to MUI v6
MUI v6 has been released. ~Need material-react-table to be updated with a stable release with support. Also ideally want to update SciGateway around the same time for which we will want react-18 to at least be merged.~ Once this is done MRT v3 will also be unblocked. There should not be an issue running SciGateway with MUI v5 and IMS with v6 but its worth checking it works correctly in case.

asuresh-code:
## Description

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #915


src/admin/units/unitsDialog.component.tsx
joelvdavies, line 1: If you run 'yarn lint' locally this file has a slight change to it, could you commit this please.

src/catalogue/items/catalogueItemsTable.component.tsx
joelvdavies, line 1: I think this file was changed by a codemod - if you save it again locally (assuming format on save is enabled) it will format it back again.

joelvdavies: I think its the same on some of the other changes in other files e.g.
```typescript
      size={{
        xs: 12,
        sm: "grow"
      }}>
```
Is using " instead of the usual '. I think you can just run `yarn prettier ./src --write` to format everything.

joelvdavies (commented):
Started going through, haven't found an issues yet. Just noticed the files could probably do with formatting after the code mods were applied.

src/handleIMS_APIError.test.ts
joelvdavies, line 1: The line endings changed here - nothing to address thought I would just tag Louise's comment here, we may want to add a .gitattributes in the future https://github.com/ral-facilities/datagateway/pull/1716#discussion_r2095474953. The same happened to handleIMS_APIError and others below.

src/admin/units/unitsDialog.component.tsx
asuresh-code: I ran yarn lint, but did not see any results/info from it. I can see github the props have been moved into the Sx, which reflects the deprecation changes as part of MUI v6
image

src/systems/systemItemsTable.component.tsx
```diff
@@ -180,6 +180,7 @@ export function SystemItemsTable(props: SystemItemsTableProps) {
           (itemData) =>
             ({
               item: itemData,
+
```
joelvdavies, line 183:
```suggestion
```

PR #1427, opened 2025-06-02
Add unit test, test selection

asuresh-code:
## Description

Uses vitest's `--changed` option to selectively run unit tests, by generating a dependency graph of imports and exports. Added a new script to try this option. Using the dependency graph to select tests is not perfect, and may miss some necessary tests. I would recommend that if approved, this be used for local development testing, and if it works well for some time, then integrate into the github workflows, since it will likely greatly reduce the number of tests run, reducing load on runners + time spent waiting.

## Testing instructions

Make changes to component/test files, then run the new script `yarn test:changed` to see which test files are run.

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

connect to {issue number}

### DIFF
--- a/src/paths.tsx
+++ b/src/paths.tsx
@@ -17,6 +17,6 @@ export const paths = {
   manufacturer: '/manufacturers/:manufacturer_id',
 };
 
-export type urlPathKeyType = keyof typeof paths;
+export type URLPathKeyType = keyof typeof paths;
 
 export default paths;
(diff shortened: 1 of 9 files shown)


joelvdavies (approved):
Works well for me locally. Good idea to try it first.

joshdimanteto (changes requested):

src/paths.tsx
```diff
@@ -16,3 +16,7 @@ export const paths = {
   manufacturers: '/manufacturers',
   manufacturer: '/manufacturers/:manufacturer_id',
 };
+
+export type urlPathKeyType = keyof typeof paths;
```
joelvdavies, line 20: Types should be pascal case - URLPathKeyType.

joshdimanteto (approved):

PR #1450, opened 2025-06-10
Some bug fixes for v2 #1449

### LINKED ISSUE #1446: Deleting default value in property dialogue displays undefined
Image

For an allowed values property - selecting a default then using backspace to delete replaces the contents with undefined. It only appears once, deleting the undefined leaves it empty as expected. This doesn't happen to the units or mandatory autocomplete.

### LINKED ISSUE #1447: Deleting allowed value property contents on catalogue item displays null
Similar to https://github.com/ral-facilities/inventory-management-system/issues/1446 but in the property dialogue itself for catalogue items.

Image

### LINKED ISSUE #1449: Units and usage status pages have scroll bars
This is visible on the current preprod

Image

joelvdavies:
## Description

Fixes scrollbars mentioned in #1449, was similar to https://github.com/ral-facilities/inventory-management-system/pull/1445, we were both fixing the height and allowing it to expand.

Fixes deleting the value of an allowed value property in the catalogue item/item dialogue previously displaying null #1447. This should be tested with string and number values as both were effected.

Fixes deleting the default value in property dialogue displays undefined #1446. Still not convinced on the reasoning for this happening but I have implemented a fix for it. Test with both integer and string values of the default value.

Fixes a scrollbar appearing when using the advanced section of the items dialogue and specifying a start index.

## Testing instructions

- [ ] Review code
- [ ] Test the fixes address the issues described above
- [ ] Check Actions build

## Agile board tracking

Closes #1449, Closes #1447, Closes #1446

### DIFF
--- a/src/admin/units/__snapshots__/units.component.test.tsx.snap
+++ b/src/admin/units/__snapshots__/units.component.test.tsx.snap
@@ -3,7 +3,7 @@
 exports[`Units > renders table correctly 1`] = `
 <DocumentFragment>
   <div
-    style="width: 100%; height: 100%;"
+    style="width: 100%;"
   >
     <div
       class="MuiPaper-root MuiPaper-elevation MuiPaper-rounded MuiPaper-elevation2 css-1s80a98-MuiPaper-root"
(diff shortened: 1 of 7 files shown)


asuresh-code (dismissed):
LGTM, all bugs were fixed when testing.

joelvdavies:
Ah sorry just found another one I will put here:
image

PR #1481, opened 2025-07-09
add systems types_id to systems dialog #1466

### LINKED ISSUE #1466: intergrate system types to add, edit and move dialogs
#### 1. Add Autocomplete Field for System Type in System Dialog
- [x] Add an **autocomplete input** for selecting the system type when creating or editing a system.
- [x] If a **parent system** is selected, automatically populate the system type with the parent’s type.
- [x]  disable the field if the type is inherited from the parent system.

joshdimanteto:
## Description
see #1466

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Check the select in different location in systems 
- [ ] Check the move to and copy to for systems 

## Agile board tracking

closes #1466


cypress/e2e/with_mock_data/systems.cy.ts
```diff
@@ -374,13 +385,33 @@ describe('Systems', () => {
         'A System with the same name already exists within the same parent System. Please enter a different name.'
       ).should('not.exist');
     });
+
+    it('displays error message if the system type id is not found', () => {
+      cy.visit('/systems');
+
+      cy.findByRole('button', { name: 'Add System' }).click();
+      cy.findByLabelText('Name *').type('Error Not found type');
```
joelvdavies, line 393:
```suggestion
      cy.findByLabelText('Name *').type('Error type not found');
```
Picky but I think its clearer when I read it/matches with others.

src/systems/systemDialog.component.test.tsx
```diff
@@ -27,12 +27,13 @@ describe('Systems Dialog', () => {
   };
 
   // Modifies values when given a value that is not undefined
-  const modifyValues = (values: {
+  const modifyValues = async (values: {
```
joelvdavies, line 30: It looks like this was changed to async just for type, is it worth changing the `getByRole` on the others to `find` or vice versa. Copilot is suggesting `getByRole` because the input is already expected to be in the dom when the function is called and so will be slightly more efficient/faster in this case.

src/systems/systemDirectoryDialog.component.tsx
```diff
@@ -177,6 +184,9 @@ export const SystemDirectoryDialog = (props: SystemDirectoryDialogProps) => {
           systemParentId={parentSystemId ?? undefined}
           onChangeParentId={setParentSystemId}
           selectedSystems={selectedSystems}
+          isSystemSelectable={(system: System) => {
+            return parentSystemTypeId === system.type_id;
+          }}
```
joelvdavies, line 189:
```suggestion
          isSystemSelectable={(system: System) => parentSystemTypeId === system.type_id }
```

joelvdavies (commented):
Just some minor comments, worked well for me 🥳

joelvdavies (approved):

PR #1497, opened 2025-08-12
Apply rules in item dialog

### LINKED ISSUE #1469: Modify items dialog to use creations and moving rules
This issue tracks the changes needed to enforce creation rules when adding new items or moving existing, specifically by reordering the dialog steps and applying system type constraints.

---

#### 1. Reorder Steps in the Item Creation Dialog
- [ ] Move the **Placement** step to be the **first step** in the item creation dialog.
- [ ] Ensure the selected system is stored and passed to subsequent steps.

#### 2. Apply System Type Rules in Placement Step
- [ ] Apply the rules to navigation view so this would be all rules that have a initial systems type X 
- [ ] Disable that do not match the allowed type.

#### 3. Auto-populate Usage Status Based on System Type
- [ ] After selecting a system in the Placement step, automatically **populate the usage status** field in the next step based on the selected system’s type.

joshdimanteto:
## Description

see #1469

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #1469

### DIFF
--- a/cypress/e2e/with_api/items/functions.ts
+++ b/cypress/e2e/with_api/items/functions.ts
@@ -169,6 +169,7 @@ export const editProperty = () => {
   cy.findByRole('button', { name: 'items landing page actions menu' }).click();
   cy.findByText('Edit').click();
   cy.findByRole('button', { name: 'Next' }).click();
+  cy.findByRole('button', { name: 'Next' }).click();
   cy.findByLabelText('Type *').click();
   cy.findByRole('option', { name: 'flat' }).click();
   cy.findByDisplayValue('flat').should('exist');
--- a/cypress/e2e/with_api/systems/functions.ts
+++ b/cypress/e2e/with_api/systems/functions.ts
@@ -170,7 +170,7 @@ export const moveItemToSystem = (values: {
   cy.findByRole('progressbar').should('not.exist');
 
   cy.findAllByRole('combobox').eq(1).click();
-  cy.findByText('Scrapped').click();
+  cy.findByText('In Use').click();
 
   cy.findByRole('button', { name: 'Finish' }).should('not.be.disabled');
   cy.findByRole('button', { name: 'Finish' }).click();


src/items/itemDialog.component.tsx
```diff
@@ -175,8 +181,26 @@ function ItemDialog(props: ItemDialogProps) {
...
+  const { data: dstSystem } = useGetSystem(parentSystemId);
+  const { data: srcSystem } = useGetSystem(selectedItem?.system_id);
+  const srcSystemTypeId =
+    requestType === 'post' ? 'null' : (srcSystem?.type_id ?? 'null');
+
+  const dstSystemTypeId = dstSystem?.type_id ?? 'null';
+  const { data: tableRules } = useGetRules(srcSystemTypeId);
+
+  // This should be a list of 1 rule
+  const { data: SelectedRule } = useGetRules(srcSystemTypeId, dstSystemTypeId);
```
joelvdavies, line 194:
```suggestion
  const { data: selectedRule } = useGetRules(srcSystemTypeId, dstSystemTypeId);
```

cypress/e2e/with_mock_data/items.cy.ts
```diff
@@ -103,14 +103,21 @@ describe('Items', () => {
...
+    cy.findByRole('progressbar').should('not.exist');
+
+    cy.findAllByText('Storage').first().click();
+
+    cy.findByText('No systems found').should('exist');
+
+    cy.findAllByText('Storage').should('have.length', 1);
+
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    cy.wait(100);
```
joelvdavies, line 115: Looks like this was needed in these tests because at the point it is trying to click next afterwards, the rules request hasn't finished so it errors as it triggers `!isDstSystemTypeSameAsSrcSystemType && (!SelectedRule || SelectedRule.length === 0)`, and even waiting for the request to finish wont necessarily be enough if it hasn't re-rendered the dialog yet.

A user would have to be very fast/the backend would have to get stuck handling a request for this to break for a user so I am not too concerned. I think the only long term solution I can think of now is having a loading spinner while it is still loading the relevant data for the form, so the tests can then wait for something to be displayed to know everything is ready.

src/form.schemas.tsx
```diff
@@ -540,7 +540,8 @@ export const ItemDetailsStepSchema = (requestType: RequestType) => {
     purchase_order_number: OptionalOrNullableStringSchema({ requestType }),
     is_defective: MandatoryBooleanSchema({}),
     usage_status_id: MandatoryStringSchema({
-      errorMessage: 'Please select a usage status.',
+      errorMessage:
+        'Please navigate back and select a system. This field will then be prepopulated.',
```
MatteoGuarnaccia5, line 544:
```suggestion
        'Please navigate back and select a system. This field will then be auto-filled.',
```

src/items/itemDialog.component.tsx
```diff
@@ -175,8 +181,26 @@ function ItemDialog(props: ItemDialogProps) {
...
+  const { data: dstSystem } = useGetSystem(parentSystemId);
+  const { data: srcSystem } = useGetSystem(selectedItem?.system_id);
+  const srcSystemTypeId =
+    requestType === 'post' ? 'null' : (srcSystem?.type_id ?? 'null');
+
+  const dstSystemTypeId = dstSystem?.type_id ?? 'null';
+  const { data: tableRules } = useGetRules(srcSystemTypeId);
+
+  // This should be a list of 1 rule
+  const { data: selectedRule } = useGetRules(srcSystemTypeId, dstSystemTypeId);
```
MatteoGuarnaccia5, line 194: Should this assumption not be handled?

src/items/itemDialog.component.tsx
```diff
@@ -175,8 +181,26 @@ function ItemDialog(props: ItemDialogProps) {
...
+
+  const isDstSystemTypeSameAsSrcSystemType = React.useMemo(() => {
+    if (!dstSystem || !srcSystem) return false;
+    return dstSystemTypeId === srcSystemTypeId;
```
MatteoGuarnaccia5, line 198:
```suggestion
    return (dstSystem && srcSystem) && (dstSystemTypeId === srcSystemTypeId);

```

joshdimanteto: This code suggestion does not read well. I'm going to stick with the current implementation

src/form.schemas.tsx
joshdimanteto: Prepopulated is fine here. No need to change unless users report confusion.

src/items/itemDialog.component.tsx
joshdimanteto: I don't know what you mean be this ? Have you look at a where `selectedRule` has been used (lines 273 - 285)?

MatteoGuarnaccia5: I meant what if it returns a list with more than a single element, to be fair the backend covers that case doesn't ?

MatteoGuarnaccia5 (approved):

src/items/itemDialog.component.tsx
joshdimanteto: it still uses the first element in the array, so If the array is greater than one than the problem would have to solved using the backend.

cypress/e2e/with_mock_data/items.cy.ts
joshdimanteto: #1517

PR #1499, opened 2025-08-14
Apply rules to move items dialog #1470

### LINKED ISSUE #1470: Enforce Rules to MoveTo Dialog
This issue tracks improvements to the **MoveTo** dialog to enforce system type movement rules and provide a clear confirmation step before completing a move.

---

#### 1. Enforce System Type Movement Rules
- [x] Apply all relevant **system type rules** when selecting a target system in the MoveTo dialog.
- [ ] Disable or hide systems that are **not valid move targets** based on the source system’s type and the item’s catalogue rules.
- [x] Ensure the UI clearly communicates why certain systems are unavailable.

#### 2. Handle One-to-One Usage Status Mapping
- [x] If the move results in a **one-to-one mapping** of usage status (e.g. moving from a Storage system to an Operational system), automatically update the usage status of the items.
- [x] Prevent manual override of usage status if the mapping is strict. (removing step 2 as that allows he user to manually change the usage status)

joshdimanteto:
## Description
see #1470

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes #1470


joshdimanteto:
#1517

cypress/e2e/with_api/systems/functions.ts
```diff
@@ -160,26 +160,17 @@ export const moveItemToSystem = (values: {
...
-  cy.findByRole('progressbar').should('not.exist');
-
-  cy.findAllByRole('combobox').eq(1).click();
-  cy.findByText('In Use').click();
-
-  cy.findByRole('button', { name: 'Finish' }).should('not.be.disabled');
-  cy.findByRole('button', { name: 'Finish' }).click();
-
+  // eslint-disable-next-line cypress/no-unnecessary-waiting
+  cy.wait(4000);
```
joshdimanteto, line 164: #1517

src/systems/systemItemsDialog.component.test.tsx
```diff
@@ -154,150 +149,77 @@ describe('SystemItemsDialog', () => {
...
-        screen.queryByText(
-          'Move items from current location or root to another system'
-        )
-      ).not.toBeInTheDocument();
+      expect(screen.queryByText(errorMessage)).not.toBeInTheDocument();
     });
 
-    it('sets the finish button to disabled when the usage statuses and system is not set and clears the states when resolved', async () => {
-      props.parentSystemId = '65328f34a40ff5301575a4e3';
+    it('cannot move selected items to root and resets back to an non error state after error have been resolved', async () => {
```
MatteoGuarnaccia5:
```suggestion
    it('cannot move selected items to root and resets back to an non error state after error has been resolved', async () => {
```

MatteoGuarnaccia5 (changes requested):
Just a single typo but otherwise looks good

joelvdavies (changes requested):
One thing I missed on the last pr and only noticed now, we originally said that we would always allow transfers between two systems of the same type while keeping the usage status the same - its 13 on the user stories. Here I cant move an item from one storage system to another because there is no rule, but I think it should be allowed in this case.

joelvdavies (approved):
Working well for me.

PR #1501, opened 2025-08-14
Apply rules to the dialog dialog #1500

### LINKED ISSUE #1500: Apply rules to delete items dialog
Add logic using the rules to check if the item is allowed to be deleted

joshdimanteto:
## Description

applying the deletion rules to the delete dialog 

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #1500


src/items/deleteItemDialog.component.tsx
```diff
@@ -16,35 +16,49 @@ import React from 'react';
...
-    if (item) {
+    const allowedSystemTypes: string[] =
+      deletionRules
+        ?.map((rule) => rule.src_system_type?.value ?? '')
+        .filter((value): value is string => value !== '') || [];
+
+    if (SelectedRule && SelectedRule.length > 0) {
       deleteItem(item)
         .then(() => {
           onClose();
```
MatteoGuarnaccia5, line 64: This logic does not make sense to me - if a rule ( any rule ) exists for the system the item can be deleted? Otherwise it cannot. @joelvdavies  how are rules implemented for deleting items? I spoke to Viktor but [the issue on the backend](https://github.com/ral-facilities/inventory-management-system-api/issues/540) doesn't make much sense either. Surely there has got to be some kind of other check about what the rule contains

src/items/deleteItemDialog.component.tsx
```diff
@@ -16,35 +16,49 @@ import React from 'react';
...
   );
 
   const { data: systemData } = useGetSystem(item?.system_id);
+
+  // This should be a list of 1 rule
+  const { data: SelectedRule } = useGetRules(
+    systemData?.type_id ?? 'null',
+    'null'
+  );
+
```
MatteoGuarnaccia5, line 44: Similar sort of thing in #1497 , why do we assume it is a single rule? 

Can a rule not be created for moving items, so i.e. 

| intitial_system_type | final_system_type | final_usage_status |
| -------- | ------- | ------- |
| storage | operational | in_use |

but then how do we know this same rule applies for deleting the items within that system? Would we not need a second rule for that, which therefore means it won't be a list of 1 rule, which also then links to my comment above

src/items/deleteItemDialog.component.tsx
joelvdavies: Yes this is right. The idea is without any deletion rule, nothing is allowed to be deleted. But a deletion rule itself looks like

| src_system_type | dst_system_type | dst_usage_status |
| -------- | ------- | ------- |
| storage | null | null |

You can think of it like, this allows an item to move from a `storage` system to a `null` system, I.e. one that doesn't exist (so its deleting). The null in the usage status is just because it doesn't matter what the usage status is. 

In this case the
```typescript
  const { data: SelectedRule } = useGetRules(
    systemData?.type_id ?? 'null',
    'null'
  );
```
Is obtaining the deletion rules only, i.e. the ones with a dst_system_type_id of `null`. (Moving rules always have this, so do creation rules). I think the name here should be `deletionRules` or something though to make it clearer.

src/items/deleteItemDialog.component.tsx
joelvdavies: The string value of `null` tells the backend to filter by `dst_system_type_id: None`. If it was sending back undefined, i.e. it wasn't added then it would not filter by this at all. So in this case it is only one, as we will only allow one deletion rule that has a `dst_system_type_id` of None in the database. Its not the same as saying match on any dst_system_type_id.

MatteoGuarnaccia5 (approved):
I do think some clear documentation should be made in regards to rules, and how they operate. Perhaps on the wiki page?

cypress/e2e/with_mock_data/items.cy.ts
```diff
@@ -1578,7 +1578,7 @@ describe('Items', () => {
     });
   });
 
-  it('deletes an item', () => {
+  it('displays error message user tries delete item from a system type which is not allowed', () => {
```
joelvdavies:
```suggestion
  it('displays error message when a user tries delete an item from a system type which is not allowed', () => {
```

src/items/deleteItemDialog.component.test.tsx
```diff
@@ -143,6 +134,20 @@ describe('delete item dialog', () => {
     expect(handleIMS_APIError).toHaveBeenCalled();
     expect(onClose).not.toHaveBeenCalled();
   });
+
+  it('displays error message user tries delete item from a system type which is not allowed', async () => {
```
joelvdavies:
```suggestion
  it('displays error message when user tries delete item from a system type which is not allowed', async () => {
```

src/items/deleteItemDialog.component.tsx
```diff
@@ -16,35 +16,49 @@ import React from 'react';
...
 
-  const [error, setError] = React.useState(false);
   const [errorMessage, setErrorMessage] = React.useState<string | undefined>(
     undefined
   );
 
   const { data: systemData } = useGetSystem(item?.system_id);
+
+  // This should be a list of 1 rule
+  const { data: SelectedRule } = useGetRules(
```
joelvdavies:
```suggestion
  const { data: deleteRules } = useGetRules(
```
Just a bit more specific, while it shouldnt be multiple maybe plural as well as its not really enforced here and should work regardless?

joshdimanteto: `deletionRules` already exist as variable. This is all the delete rules. I'll change `SelectedRule` to `selectedRules` so it is in the right case (camelcase)

joshdimanteto: I'll also update it to be plural everywhere

joelvdavies: Ah yeah I see it was on the next line 🤦‍♂️. Thanks.

src/items/deleteItemDialog.component.test.tsx
joelvdavies: Sorry 🤦‍♂️. This still didn't make sense 😅
```
displays error message when a user tries delete an item from a system type which is not allowed
```

joelvdavies (commented):
Thanks for addressing the comments, just my own suggestion that was bad 😄

MatteoGuarnaccia5 (dismissed):

joelvdavies (approved):

PR #1502, opened 2025-08-18
Add "Spares" Column to System Type Table #1475

### LINKED ISSUE #1475: Add "Spares" Column to System Type Table
- [x] Add a new column called **"Spares"** to the system type table.
- [x] If the system type is included in the backend-defined **spares definition**, display **"Yes"**; otherwise, display **"No"**.

joshdimanteto:
## Description

see #1475

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #1475

### DIFF
--- a/src/admin/systemTypes/systemTypes.component.tsx
+++ b/src/admin/systemTypes/systemTypes.component.tsx
@@ -11,6 +11,7 @@ import type { SystemType } from '../../api/api.types';
 import { useGetSparesDefinition } from '../../api/settings';
 import { useGetSystemTypes } from '../../api/systems';
 import { usePreservedTableState } from '../../common/preservedTableState.component';
+
 import {
   COLUMN_FILTER_BOOLEAN_OPTIONS,
   COLUMN_FILTER_FUNCTIONS,
@@ -31,7 +32,7 @@ interface TableRowData extends SystemType {
 }
 
 function SystemTypes() {
-  const { data: systemTypesData, isLoading: isLoadingSystemTypes } =
+  const { data: systemTypesData = [], isLoading: isLoadingSystemTypes } =
     useGetSystemTypes();
 
   const { data: sparesDefinition, isLoading: isLoadingSparesDefinition } =
@@ -46,9 +47,9 @@ function SystemTypes() {
       setTableRows(
         systemTypesData.map((type) => ({
           ...type,
-          isSpare:
-            sparesDefinition?.system_types.some((def) => def.id === type.id) ??
-            false,
+          isSpare: !sparesDefinition
+            ? false
+            : sparesDefinition.system_types.some((def) => def.id === type.id),
         }))
       );
     }
@@ -70,7 +71,7 @@ function SystemTypes() {
         Cell: ({ row }) => row.original.value,
       },
       {
-        header: 'Is Spare',
+        header: 'Used For Spares',
         Header: TableHeaderOverflowTip,
         accessorFn: (row) => (row.isSpare === true ? 'Yes' : 'No'),
         id: 'isSpare',
@@ -181,8 +182,8 @@ function SystemTypes() {
           const actualFilter = actualFilters.find((f) => f.id === id);
           if (!actualFilter) return false;
           if (actualFilterFns[id] !== filterFn) return false;
-          // Compare values stringified (arrays)
 
+          // Compare values stringified (arrays)
           return JSON.stringify(actualFilter.value) === JSON.stringify(value);
         });
       }
@@ -190,7 +191,7 @@ function SystemTypes() {
         <Box>
           <Button
             startIcon={<ClearIcon />}
-            sx={{ mx: '4px' }}
+            sx={{ mx: 0.5 }}
             variant="outlined"
             disabled={preservedState.columnFilters.length === 0}
             onClick={() => {
@@ -219,7 +220,7 @@ function SystemTypes() {
               ]);
             }}
           >
-            Spares Definition
+            Show Spares Definition
           </Button>
         </Box>
       );
--- a/src/admin/systemTypes/systemsTypes.component.test.tsx
+++ b/src/admin/systemTypes/systemsTypes.component.test.tsx
@@ -59,7 +59,7 @@ describe('SystemTypes', () => {
     expect(clearFiltersButton).toBeDisabled();
   }, 10000);
 
-  it('sets and clears the spares definition', async () => {
+  it('sets and clears the spares definition filter', async () => {
     createView();
 
     await waitFor(() => {
@@ -67,7 +67,7 @@ describe('SystemTypes', () => {
     });
 
     const sparesDefinitionButton = screen.getByRole('button', {
-      name: 'Spares Definition',
+      name: 'Show Spares Definition',
     });
 
     expect(sparesDefinitionButton).not.toBeDisabled();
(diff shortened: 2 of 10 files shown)


src/admin/systemTypes/systemTypes.component.tsx
```diff
@@ -130,21 +166,64 @@ function SystemTypes() {
...
+              table.resetGlobalFilter();
+              table.setColumnFilters([
+                {
+                  id: 'isSpare',
+                  value: COLUMN_FILTER_BOOLEAN_OPTIONS[0],
+                },
+              ]);
+            }}
+          >
+            Spares Definition
```
joelvdavies:
```suggestion
            View spares definition
```
I don't think its clear what the button is for without saying view?

src/admin/systemTypes/systemTypes.component.tsx
```diff
@@ -43,6 +69,16 @@ function SystemTypes() {
         columnFilterModeOptions: COLUMN_FILTER_MODE_OPTIONS.string,
         Cell: ({ row }) => row.original.value,
       },
+      {
+        header: 'Is Spare',
```
joelvdavies:
```suggestion
        header: 'Contains spares',
```
I feel emphasising the systems with the type contains them is clearer?

src/api/settings.tsx
```diff
@@ -0,0 +1,22 @@
+import { useQuery, UseQueryResult } from '@tanstack/react-query';
+import { AxiosError } from 'axios';
+import { imsApi } from './api';
+import { SparesDefinition } from './api.types';
+
+const getSparesDefinition = async (): Promise<SparesDefinition> => {
+  return imsApi.get('/v1/settings/spares-definition').then((response) => {
+    return response.data;
+  });
+};
```
joelvdavies: This can return a 204 when there is no spares definition defined. Right now when I go to view system types I get

Image

when it hasn't been. I know we are not expecting them to use it without, but if its an easy change it would be nice to allow for it for the time being since we aren't making it compulsory in the backend just yet. I dont think its necessarily worth adding an e2e test though.

src/admin/systemTypes/systemTypes.component.tsx
```diff
@@ -130,21 +166,64 @@ function SystemTypes() {
...
+          // Compare values stringified (arrays)
+
+          return JSON.stringify(actualFilter.value) === JSON.stringify(value);
+        });
+      }
+      return (
+        <Box>
+          <Button
+            startIcon={<ClearIcon />}
+            sx={{ mx: '4px' }}
```
joelvdavies:
```suggestion
            sx={{ mx: 0.5 }}
```
To match below.

src/admin/systemTypes/systemTypes.component.tsx
```diff
@@ -130,21 +166,64 @@ function SystemTypes() {
...
+        if (actualFilters.length !== expectedFilters.length) return false;
+
+        // Check every expected filter matches actual filter and filterFn
+        return expectedFilters.every(({ id, filterFn, value }) => {
+          const actualFilter = actualFilters.find((f) => f.id === id);
+          if (!actualFilter) return false;
+          if (actualFilterFns[id] !== filterFn) return false;
+          // Compare values stringified (arrays)
+
+          return JSON.stringify(actualFilter.value) === JSON.stringify(value);
```
joelvdavies:
```suggestion

          // Compare values stringified (arrays)
          return JSON.stringify(actualFilter.value) === JSON.stringify(value);
```

src/admin/systemTypes/systemsTypes.component.test.tsx
```diff
@@ -58,4 +58,38 @@ describe('SystemTypes', () => {
 
     expect(clearFiltersButton).toBeDisabled();
   }, 10000);
+
+  it('sets and clears the spares definition', async () => {
```
joelvdavies: Suggested change
      
    
    
      
          
            
              it('sets and clears the spares definition', async () => {
          
          
            
              it('views and clears the spares definition', async () => {
          
      
    
    
  

It is sets in the context of the filters though? If modified here same for e2e with mock data.

src/admin/adminCardView.component.tsx
```diff
@@ -1,13 +1,28 @@
...
   return (
     <Grid container>
       <AdminCard title="Units" to="units" />
       <AdminCard title="Usage Statuses" to="usage-statuses" />
       <AdminCard title="System Types" to="system-types" />
       <AdminCard title="Rules" to="rules" />
+      <AdminCard
+        title="Spares Definition"
+        to={`system-types${sparesFilterState}`}
+      />
```
MatteoGuarnaccia5, line 25: The wording of `Spares definition` is confusing. It does not clearly explain you are just looking at system types which can have spares? That is my understanding of what spares are, but it is quite confusing in the way it is presented - some of Joels comments also allude to this though.

Also the `Spare definition` card taking you to the `System Type` table but just filtered to me makes it redundant, and again adds confusion as it implies it is a completely separate thing or that it is going to lead to a definition/explanation of what a spare is given the name of the card.

src/admin/systemTypes/systemTypes.component.tsx
```diff
@@ -24,14 +26,38 @@ import {
...
+      setTableRows(
+        systemTypesData.map((type) => ({
+          ...type,
+          isSpare:
+            sparesDefinition?.system_types.some((def) => def.id === type.id) ??
+            false,
+        }))
+      );
+    }
+  }, [systemTypesData, isLoading, sparesDefinition]);
```
MatteoGuarnaccia5, line 56: image

I am getting this error quite often - given it may be just as I am using mock data and on dev mode.

joelvdavies: Do you meant the mock data from msw, or the one from the backend. If the latter see my above comment, I think its due to the spares definition not being set. You probably need to run `python ./scripts/dev_cli.py db-import` again to update the data (assuming you are happy with it being replaced).

MatteoGuarnaccia5: I meant from msw

src/admin/systemTypes/systemTypes.component.tsx
joshdimanteto: I agree with changing the column name, however `Contains spares` is misleading because it suggests the system type might include other kinds of items as well.

I asked Copilot for some suggestions:

```
Spare-Only Type  
Allows Only Spare Items  
Restricted to Spares  
Spare Item Type  
Spare-Exclusive System Type  
Spare Items Only  
Is Spare-Only?  
Spare Type (Yes/No)  
System Type for Spares  
Spare Use Only
```

I like the last two. what is you preference from the list?

joelvdavies: Yeah out of those I lean towards the last two. Perhaps a slight rephrasing e.g. "Used for spares"?

src/admin/systemTypes/systemTypes.component.tsx
joshdimanteto: I will use `Show` instead of `View` as this is used in the Rules table. 
image

src/admin/adminCardView.component.tsx
joshdimanteto: The “Spares definition” card is just meant to take you to the System Type table so you can see which system types are marked as spares. It’s mainly for quick access. The idea is that users can go to the page to understand that if an item belongs to a system with system type X, then it’s considered a spare. If that’s not clear, they can use the help section for more detail.

PR #1504, opened 2025-08-18
Spares Integration for Catalogue Item Landing Page #1477

### LINKED ISSUE #1477: Spares Integration for Catalogue Item Landing Page
- [x] Display the **number of spares** for the current catalogue item.
- [x] Make the count **clickable**, linking to the **Items Table** with the **spares filter applied** for that specific catalogue item.

joshdimanteto:
## Description

see #1477

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #1477

### DIFF
--- a/src/catalogue/items/catalogueItemsLandingPage.component.test.tsx
+++ b/src/catalogue/items/catalogueItemsLandingPage.component.test.tsx
@@ -227,7 +227,7 @@ describe('Catalogue Items Landing Page', () => {
     expect(url).toHaveAttribute('href', '/catalogue/5/items/89/items');
   });
 
-  it('navigates to items table view (spares definition)', async () => {
+  it('navigates to items page with spares definition applied', async () => {
     createView('/catalogue/5/items/89');
     await waitFor(() => {
       expect(screen.getByRole('link', { name: '0' })).toBeInTheDocument();


cypress/e2e/with_api/systems/functions.ts
```diff
@@ -214,6 +214,18 @@ export const navigateToItemsTableViaSpares = () => {
   cy.findByText('Actions').should('exist');
   cy.findByRole('link', { name: 'Spherical Lenses' }).click();
   cy.findByText('Number of spares').should('exist');
+  // eslint-disable-next-line cypress/no-unnecessary-waiting
+  cy.wait(1000);
```
MatteoGuarnaccia5, line 218: Just curiosity - I've seen more and more of these cy.wait - I assume to rendering times? Is it a workaround for an issue and something we plan on fixing, or just how we have to do it.

joshdimanteto: This due to the cypress being to fast. If this is use in dev/prod mode this is not an issue.

joelvdavies: Yeah its because in some cases we dont have anything to wait on currently, e.g. the form for moving systems has to fetch the rules, but without a loading indicator (which we dont have as it would always be a very fast query, so just flickers) we have nothing to really wait for. Thinking about it we probably could wait for the network request though, but even then it would have to re-render before its ready to proceed, I think it might work but not sure.

src/catalogue/items/catalogueItemsLandingPage.component.test.tsx
```diff
@@ -227,6 +227,21 @@ describe('Catalogue Items Landing Page', () => {
     expect(url).toHaveAttribute('href', '/catalogue/5/items/89/items');
   });
 
+  it('navigates to items table view (spares definition)', async () => {
```
MatteoGuarnaccia5:
```suggestion
  it('navigates to items table view (spares definition applied)', async () => {
```

src/catalogue/items/catalogueItemsLandingPage.component.test.tsx
```diff
@@ -227,6 +227,21 @@ describe('Catalogue Items Landing Page', () => {
...
+
+    const url = screen.getByRole('link', {
+      name: '0',
+    });
+    expect(url).toHaveAttribute(
+      'href',
+      '/catalogue/5/items/89/items?state=N4IgxgYiBcDaoEsAmMQGcCeaAuBTAtgHTYYAOuhAbgIYA2ArriADQg0NNygnmo4BOCAHYBzFmzqNUAZWwB7ftRFMAvgF11KoA'
+    );
+  });
+
```
MatteoGuarnaccia5, line 244: Is it possible to also check the table has rendered it correctly? Rather than just checking the url has the right state

src/catalogue/items/catalogueItemsLandingPage.component.test.tsx
joelvdavies: The e2e mock data test even uses `navigates to items page with spares definition applied`, probably worth keeping them consistent.

src/catalogue/items/catalogueItemsLandingPage.component.test.tsx
joshdimanteto: There is no need to test this as this is test in the e2e tests and the items table itself

joshdimanteto: The test you suggested isn’t possible as a unit test because of how the routing is implemented. The test only renders the LandingPage component and no other components.

joshdimanteto:
> Just thinking from yesterday, the refresh button is probably needed with this work on spares, so that users can refresh and see the most accurate number of spares for example.

It is not essential for now as the user can just manually refresh the page, if we have time after all the priority V3 we can look into again.

MatteoGuarnaccia5 (approved):

PR #1505, opened 2025-08-19
Spares intergration items table #1478

### LINKED ISSUE #1478: Spares Integration for Items Table
- [x] When clicked:
  - Fetch the **spares definition** from the backend.
  - Apply the appropriate **filtering logic** (e.g. filter by system types like "Storage").
  - **Clear all other filters** to show only spare items.

joshdimanteto:
## Description

see #1478

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #1478

### DIFF
--- a/src/utils.tsx
+++ b/src/utils.tsx
@@ -607,10 +607,12 @@ export const useSparesFilterState = (
   } = useGetSparesDefinition();
 
   const urlParam = urlParamName || 'state';
-  const sparesFilter = sparesDefinition.system_types.map((type) => ({
-    type: 'string',
-    value: type.value,
-  }));
+  const sparesFilter = sparesDefinition
+    ? sparesDefinition.system_types.map((type) => ({
+        type: 'string',
+        value: type.value,
+      }))
+    : false;
 
   const sparesColumnsFilters: { cF: MRT_ColumnFiltersState } = React.useMemo(
     () => ({
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -89,11 +89,6 @@ jobs:
           repository: ral-facilities/inventory-management-system-api
           ref: develop
 
-      - name: Set up Python
-        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
-        with:
-          python-version: '3.12'
-
       # This is required as need to setup api in a different directory as checkout will attempt delete
       # all existing files which in this case will include a data directory created by docker causing
       # a permission error (checkout action also can't specify a different directory to clone into)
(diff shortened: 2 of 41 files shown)


src/items/itemsTable.component.tsx
```diff
@@ -596,6 +608,18 @@ export function ItemsTable(props: ItemTableProps) {
...
+          sx={{ mx: 0.5 }}
+          variant="outlined"
+          disabled={searchParams.get('state') === encodedSparesFilter}
+          onClick={() => {
+            const newParams = new URLSearchParams(searchParams);
+            newParams.set('state', encodedSparesFilter);
+            setSearchParams(newParams, { replace: false });
+          }}
+        >
+          Show Spare Items
```
MatteoGuarnaccia5, line 621: Do CLF users know and understand what a `spare item` is? There is no explanation given in the table

src/utils.tsx
```diff
@@ -596,7 +596,11 @@ export const COLUMN_FILTER_BOOLEAN_OPTIONS = ['Yes', 'No'];
...
   urlParamName?: string
-): { sparesFilterState: string; isLoading: boolean } => {
+): {
+  sparesFilterState: string;
+  encodedSparesFilter: string;
+  isLoading: boolean;
+} => {
   const {
     data: sparesDefinition = { system_types: [] },
     isLoading: isLoadingSparesDefinition,
```
MatteoGuarnaccia5, line 606: The `isLoading` here only gets referenced again in the return statement, shouldn't the logic of mapping the result of `useGetSparesDefenition` account for this loading?

joshdimanteto: Can you elaborate on this please ?

src/items/itemsTable.component.tsx
joshdimanteto: For now, the help text and the user feedback session will be enough, if the users request this, then it will be addressed

src/utils.tsx
MatteoGuarnaccia5: In the rest of the function, it determines the `sparesDefinition`, without waiting for the `isLoading` to be false. It is not something you have changed in this PR, but to me should it not wait on the request to stop loading before it tries to format the spares definition

joshdimanteto: We don’t need to wait for `isLoading` to be `false` before formatting `sparesDefinition` because React (and hooks like React Query) are designed to render immediately with whatever data is available and then re-render when new data arrives. Blocking until loading finishes would prevent the component from rendering, which goes against React’s declarative approach.

In this case:

- `useGetSparesDefinition()` returns `data` and `isLoading`.
- While loading, `data` may be `undefined`, so we provide a safe default:  
  ```ts
  data: sparesDefinition = { system_types: [] }
  ```
- The function uses that default to build `sparesFilter` and other values. When the real data arrives, React triggers a re-render and updates everything automatically.

This pattern avoids conditional rendering complexity and ensures the component always has a valid state, even during loading. If we waited explicitly, we’d either block rendering or introduce unnecessary checks that React already handles for us.

joelvdavies: Yeah I think the most we would do is return undefined when it hasnt loaded yet, waiting would prevent the render, but the way this hook works is the same as the useGetSparesDefinition it uses - i.e. it returns if its finished loading or not rather than waiting for it to finish. React than handles the update automatically when the state does change re-rendering the appropriate component that uses the hook.

MatteoGuarnaccia5 (approved):

Issue #1474, opened 2025-07-07
Spares Integration Across Catalogue Items and Items Table


joshdimanteto:
This issue introduces support for displaying and interacting with **spare items** across the Catalogue Items Table, Catalogue Item Landing Page, and Items Table.

---

Issue #1160, opened 2024-12-16
Spares 


joshuadkitenge:

joelvdavies:
Closed via https://github.com/ral-facilities/inventory-management-system/pull/1509. (Link used there was to #1150 instead)

Issue #1095, opened 2024-10-28
Negative max field in tables filters incorrectly on production


joelvdavies:
Was found here https://github.com/ral-facilities/inventory-management-system/pull/1087#discussion_r1819194053. Currently traced to https://github.com/ral-facilities/inventory-management-system/pull/922 having fixed it for catalogue items and items tables. Not sure why though. It is currently present in production and may effect other tables as well.

joelvdavies:
Traced it to the preserved state changes - still not sure why, but passing columnFilterFns: {} as the table state fixes the issue.

joelvdavies:
Its also present on their MRTs website: https://www.material-react-table.com/docs/guides/column-filtering. Only when I go to open it in code sandbox the issue disappears. I tried recreating my lockfile and deleting MRT but it still did not fix.

joelvdavies:
https://github.com/ral-facilities/inventory-management-system/pull/1046 breaks it just for the items table for some reason.

joshuadkitenge:
I think this occurs as MRT assume the that the default minimum value is `0 ` so if you have a max range to `-687` it would be invalid as it from `min: 0 to max: -687`.

There are really two valid fixes for this. 

1. We add extra error handling to the range columns filters in MRT (for invalid ranges)
2. We change the minimum value default value from 0 to negative infinity

joshuadkitenge:
#634

PR #1509, opened 2025-09-10
spares integration on systems item table #1483

### LINKED ISSUE #1483: Spares intergration for system items table
add column with the spares number for each item

joshdimanteto:
## Description

see #1483

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #1483


src/systems/systemItemsTable.component.tsx
```diff
@@ -228,6 +233,40 @@ export function SystemItemsTable(props: SystemItemsTableProps) {
...
+        // If not, it does not display. This seems to be a Material React Table (MRT) issue.
+        accessorFn: (row) => String(row.catalogueItem.number_of_spares ?? 0),
+        id: 'catalogueItem.number_of_spares',
+        filterVariant: COLUMN_FILTER_VARIANTS.number,
+        filterFn: COLUMN_FILTER_FUNCTIONS.number,
+        columnFilterModeOptions: [
+          ...COLUMN_FILTER_MODE_OPTIONS.number,
+          ...OPTIONAL_FILTER_MODE_OPTIONS,
+        ],
+        GroupedCell: TableGroupedCell,
```
joelvdavies, line 249: Should this also be a link?

Image

cypress/e2e/with_mock_data/systems.cy.ts
```diff
@@ -298,6 +298,29 @@ describe('Systems', () => {
     cy.findByText('Properties').should('be.visible');
   });
 
+  it("should be able to navigate to an filtered item's table using the spares value", () => {
```
joelvdavies:
```suggestion
  it("should be able to navigate to a filtered item's table using the spares value", () => {
```

cypress/e2e/with_mock_data/systems.cy.ts
```diff
@@ -298,6 +298,29 @@ describe('Systems', () => {
...
+
+    // Wait for progress bar to disappear before interacting with filters
+    cy.findAllByRole('progressbar', { timeout: 10000 }).should('not.exist');
+
+    cy.findAllByRole('button', { name: 'Expand' }).eq(1).scrollIntoView();
+    cy.findAllByRole('button', { name: 'Expand' }).eq(1).click();
+    cy.findAllByRole('link', { name: '0' }).eq(2).scrollIntoView();
+    cy.findAllByRole('link', { name: '0' }).eq(2).click({ force: true });
+
+    // Check now on landing page for the item
```
joelvdavies:
```suggestion
    // Check now on table view for the item
```

src/systems/systemItemsTable.component.tsx
joshdimanteto: I don't think it can as multiple different items with different catalogue item can have same spares number.

for example 

Item 1 (Catalogue item 1 ) with spares 1
Item 2 (Catalogue item 2 ) with spares 1

if you grouped by the number of spares it would include both

joelvdavies: Oh yeah 🤦‍♂️ nevermind, thanks

joelvdavies:
> image
> Also I know it is stupid but putting a negative number in the max still displays results



I think this is probably another case for https://github.com/ral-facilities/inventory-management-system/issues/1095, good spot.

joshdimanteto:
> image
> 
> The `Number of Spares` column being there to me makes it related to the item, rather than the catalogue item. Also when viewing all the items, it shows on every row.
> 
> I think it would be better placed just to the right of the `Catalogue Item` , so it also doesn't show on every row

This is the intended behavior, as we do the same thing with the catalogue item name and in future the criticality would be implemented similar. The table is grouped by default however when ungroup we must preserve the data for the user for each item which includes catalogue item name , spares etc

joelvdavies (approved):

joelvdavies:
Will this also close https://github.com/ral-facilities/inventory-management-system/issues/1474 & https://github.com/ral-facilities/inventory-management-system/issues/1160 ? Might be worth linking them in the description before merging if so.

Issue #1472, opened 2025-07-07
Split Admin Page into "Settings" and "Admin Settings"


joshdimanteto:
This issue proposes splitting the current Admin page into two separate views: **Settings** and **Admin Settings**, to improve clarity and access control.

---

#### 1. Create a New "Settings" Page
- [ ] Display the following tables that are accessible to all users:
  - **Units**
  - **Usage Statuses**
  - **System Types**
  - **Spares definition**
  - **System Rules**
- [ ] Ensure all tables are **read-only** for non-admin users.

#### 2. Create a Separate "Admin Settings" Page
- [ ] Use the same layout and structure as the current Admin page.
- [ ] Allow full **add/edit/delete** functionality for:
  - **Units**
  - **Usage Statuses**
- [ ] Restrict access to this page to users with admin privileges.

#### 3. Add paths 
- [ ] Rename admin to settings and change the route path
- [ ] Create a route for admin settings, this should begin with `/admin/`. This page can only be view in the sci gateway admin page when run in build mode

PR #1518, opened 2025-09-24
Refactor admin page to utilise authorisation #1472

MatteoGuarnaccia5:
## Description

Based off original issue #1472, admin functionality is added via `parseToken.ts` , which will return if the current user's token has any role that is also a part of the configured `privilegedRoles` . If the user does have a privileged status, they will be able to edit/create units/usage statuses (completed in this PR), bypass rules (to be done #1523 ), and an alert with their role will appear in the top right of the website.

The `/admin` route, along with component and filenames have been renamed to be `/settings` , `Settings` etc

Please note:
This PR is dependent on https://github.com/ral-facilities/ldap-jwt-auth/pull/271 in ldap-jwt-auth, and should be merged in conjunction with it, to avoid breaking changes.

This PR adds a new key/value pair to the `inventory-management-system-settings.json` , this change should also be made in scigateway for production.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

connect to #1472


README.md
```diff
@@ -127,3 +127,7 @@ This guide outlines the steps to perform end-to-end testing for the Inventory Ma
 By following these instructions, you can effectively conduct end-to-end testing for the Inventory Management System. Choose the method that best suits your testing requirements and development environment.
 
 **Note:** On the CI environment, the IMS-API is accessed via `127.0.0.1` due to issues with `localhost`.
+
+#### Authorisation
+
+The current implementation of authorisation utilises an `AuthProvider` component to check if the user is an admin or not. A user is deemed an admin if the role within their access token is one of the `privilegedRoles` defined in `inventory-management-system-settings.json`. Currently this is just `admin`, therefore the boolean variables used across the repository to signify admin functionality are named `isAdminUser`.
```
joshdimanteto: This section should explain how to set up authorisation (for dev or SciGateway) and also outline what extra functionality admins have, rather than just describing the internal code.

cypress/e2e/with_api/catalogueCategories/catalogueCategories.cy.ts
```diff
@@ -10,6 +10,12 @@ import {
 
 describe('Catalogue Category', () => {
   beforeEach(() => {
+    cy.window().then((win) => {
+      win.localStorage.setItem(
+        'scigateway:token',
+        'eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VybmFtZSI6InVzZXJuYW1lIiwicm9sZSI6ImFkbWluIiwidXNlcklzQWRtaW4iOmZhbHNlLCJleHAiOjI1MzQwMjMwMDc5OX0.FrsDUqnKskhIvmIjtYVgC9im-cSu1dFlwVQ4cFJf2BgCaSh82XuEngOLkbtQuuXWC1wiipsGP4Y-usq7Q_R68vwXqGYusHo4fXw6AcBcwplgXZ3n60wsTegpBxKZY5foOre0Ng1GpK-7rrx9H-YQUCHSBOtzWOw_eLzu-eNTwMnMnnpGM9L91_hj0dAKiP90Z3Hp0UelnYydc0sf6msOs7RKI2Sij-13vFSL8LToIbfUTZYwKZHbBPD5glce_gsW6_W5W-iGemt7yyhfyf7IxKWq3Q02HCiSkI0uCcBal44sabPrsQ4EaPRwyUnH0X25MC00IAPRHh-1KqabV7IA9w'
+      );
+    });
```
joshdimanteto: This code is reused in multiple places, are the tokens the same or are they different ?  is there any clean up for this logic or does this persistent between different test instances . Also what is this token ?

src/authProvider.component.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { getUserRole } from './parseTokens';
+import { InventoryManagementSystemSettingsContext } from './configProvider.component';
+
+const AuthContext = React.createContext<{
+  role: string;
+  isAdminUser: boolean;
```
joshdimanteto:
```suggestion
  isPrivilegedUser: boolean;
```

I had a conversation with @joelvdavies  and for now it would be best to change `isAdminUser` to `isPrivilegedUser`. As better describes the functionality. It should also be detailed that a privilege user is equivalent to an admin for the frontend in the README.

joshdimanteto (changes requested):
Can you add a setup script or function (this could live in main.tsx potentially) to make local setup easier. When dev mode is built, it should set a value in localStorage (non-admin) via (potentially) main.tsx so the user can toggle between admin and non-admin modes. The toggle should be available in the UI as a button.

cypress/e2e/with_api/catalogueCategories/catalogueCategories.cy.ts
MatteoGuarnaccia5: No it does not seem to persist between tests, I can add a comment to describe what the token is - it is simply a token with `role: admin, isUserAdmin: false`. They are the same everywhere, I think I can move it to the config perhaps - makes it more readable at least

joshdimanteto: Look at how datagateway does it

joelvdavies:
I have a scrollbar on catalogue categories when logged in as an admin:
image
It also occurrs in catalogue items and items tables.

src/common/authRoleStatus.component.tsx
```diff
@@ -0,0 +1,67 @@
...
+          justifyContent="center"
+          aria-label="admin-status-info"
+        >
+          <Grid>
+            <Tooltip
+              title={
+                <h4>
+                  As a privileged user, you can create/edit settings, and bypass
+                  moving rules for items.
+                </h4>
```
joelvdavies: Not sure on the language here - creating settings. I am not too sure about calling usage statuses and units as settings, but struggle to think of a nice alternative. On the backend we the only "setting" we refer to is the spares definition. I did think maybe configuration, but its probably worse for a user and means virtually the same thing.

As a side note I wouldnt use h4 here, best to style the tooltip itself if you want to change the contents/use a typography with variant=h4, as for accessibility the heading order is supposed to be consistent and this is a tooltip not a heading (I made the open data label before working on any accessibility 😄).

src/settings/usageStatuses/usageStatuses.component.test.tsx
```diff
@@ -10,6 +11,10 @@ describe('Usage statuses', () => {
   };
   beforeEach(() => {
     user = userEvent.setup();
+    vi.spyOn(authProvider, 'useAuthorisationState').mockReturnValue({
+      role: 'admin',
+      isPrivilegedUser: true,
+    });
```
joelvdavies, line 17: How do we know any of these pages will work correctly for a non-admin? It think it would be worth having at least one test for each page to cover normal users as I dont think we should only test the case when everyone is an admin. The mock e2e test is similar, though we dont test much there anyway so I think they could feasibly be left.

src/parseTokens.test.ts
```diff
@@ -1,34 +1,64 @@
...
-  });
+  describe('readSciGatewayToken', () => {
+    afterEach(() => {
+      vi.clearAllMocks();
+    });
+
+    it('should read token from local storage', () => {
+      localStorageGetItemMock.mockImplementationOnce(
+        () =>
+          'eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VybmFtZSI6InVzZXJuYW1lIiwicm9sZSI6ImFkbWluIiwidXNlcklzQWRtaW4iOmZhbHNlLCJleHAiOjI1MzQwMjMwMDc5OX0.FrsDUqnKskhIvmIjtYVgC9im-cSu1dFlwVQ4cFJf2BgCaSh82XuEngOLkbtQuuXWC1wiipsGP4Y-usq7Q_R68vwXqGYusHo4fXw6AcBcwplgXZ3n60wsTegpBxKZY5foOre0Ng1GpK-7rrx9H-YQUCHSBOtzWOw_eLzu-eNTwMnMnnpGM9L91_hj0dAKiP90Z3Hp0UelnYydc0sf6msOs7RKI2Sij-13vFSL8LToIbfUTZYwKZHbBPD5glce_gsW6_W5W-iGemt7yyhfyf7IxKWq3Q02HCiSkI0uCcBal44sabPrsQ4EaPRwyUnH0X25MC00IAPRHh-1KqabV7IA9w'
```
joelvdavies: Could all of these tokens into constants with a label of what they are like was done on the backend please, same for any others in the other files.

cypress/e2e/with_mock_data/settings.cy.ts
joelvdavies, line 1: I think the the name of this file should be renamed as its now the settings page.

docker/docker-entrypoint.sh
```diff
@@ -18,6 +18,7 @@ jq \
    .attachmentAllowedFileExtensions = $attachmentAllowedFileExtensions |
    .imageAllowedFileExtensions = $imageAllowedFileExtensions |
    .maxImageSizeBytes = $maxImageSizeBytes |
+   .privilegedRoles = $privilegedRoles |
```
joelvdavies, line 22: I think you would also need to define the privilegedRoles using another `--argjson` parameter above as well.

I think its best to add another ENV in the Dockerfile.prod so it takes a default value too.

joelvdavies (commented):
Working well, like the toggle on dev mode.

src/settings/usageStatuses/usageStatuses.component.test.tsx
MatteoGuarnaccia5: I've added unit tests to test non-admin, I don't think e2e tests are really needed as you are just asserting the table renders without the button / row actions - not that they work

joelvdavies:
Could we hide this when you the user is not a privileged user?
image
Also would it be possible to remove the space at the top of it too, similar to datagateway. I think the breadcrumbs could probably be centered vertically as well as it looks a bit off?

README.md
```diff
@@ -127,3 +127,14 @@ This guide outlines the steps to perform end-to-end testing for the Inventory Ma
...
+#### Authorisation
+
+To set up authorisation correctly, make sure the `privilegedRoles` field in `inventory-management-system-settings.json` is properly configured.
+
+The system checks the role associated with the access token — whether it's obtained via manual setup, the development toggle (see `baseLayoutHeader.component.tsx`), or through SciGateway authentication — against the `privilegedRoles`. This determines whether the current user is a privileged user.
+
+**Privileged users are allowed to:**
+
+- Create and delete units and usage statuses
+- Bypass rules when creating, deleting, editing, or moving items
```
joshdimanteto, line 142: The first sentence is redundant as the setup instructions for IMS frontend say to copy over the example config file where this is already configured.

What is manual setup ?? I don't think this should be an option

Why is it referencing the file of where the development toggle is,  rather than the location on the page ? 

There is no mention that privileged users are admins. In the frontend there is no plans on having any other user than admin and default with the current implementation. 

It needs to be clearer that privileged users have these permissions on top of the standard stuff. Currently it reads like they are only allowed to do these options.

src/authProvider.component.tsx
```diff
@@ -0,0 +1,45 @@
...
+        role: role,
+        isPrivilegedUser: privilegedRoles.includes(role),
+      });
+    };
+    setAuthorisationState();
+
+    // add event listener for if token in localstorage changes
+    window.addEventListener('tokenChanged', (_) => {
+      setAuthorisationState();
+    });
```
joshdimanteto: is this needed for prod ?

src/common/authRoleStatus.component.tsx
```diff
@@ -0,0 +1,65 @@
...
+          container
+          direction="row"
+          alignItems="center"
+          justifyContent="center"
+          aria-label="admin-status-info"
+        >
+          <Grid>
+            <Tooltip
+              title={
+                'As a privileged user you can create/delete units and usage statuses, and bypass rules when creating, deleting, editing, or moving items.'
```
joshdimanteto: The text should use "Admin user" rather than "Privileged user". potentially you can look at configuring it to what ever the role name is. The expectation is that that privileged roles in the settings is a list of one (for now).  


 Similar to above, you need to mention that this is on top of the standard features

src/common/authRoleStatus.component.tsx
```diff
@@ -0,0 +1,65 @@
...
+function AuthRoleStatus() {
+  const role = getUserRole();
+
+  return (
+    role && (
+      <Paper
+        square
+        sx={{
+          maxHeight: '40px',
+          backgroundColor: '#FFA500',
```
joshdimanteto: you shouldn't be statically setting colours in IMS frontend. Look at the datagateway code (open data label) to see what variables you should be using instead

src/common/authRoleStatus.component.tsx
```diff
@@ -0,0 +1,65 @@
...
+              aria-label={'admin-status-tooltip'}
+            >
+              <IconButton
+                disableRipple
+                sx={{ backgroundColor: 'transparent' }}
+                size="large"
+              >
+                <InfoOutlinedIcon
+                  sx={{
+                    color: '#003088',
```
joshdimanteto: ^^same

src/common/authRoleStatus.component.tsx
```diff
@@ -0,0 +1,65 @@
...
+                  }}
+                />
+              </IconButton>
+            </Tooltip>
+          </Grid>
+          <Grid>
+            <Typography
+              variant="h6"
+              sx={{ color: '#000000', fontSize: '16px' }}
```
joshdimanteto: ^^same

src/parseTokens.ts
```diff
@@ -28,3 +28,15 @@ export const readSciGatewayToken = (): SciGatewayToken => {
...
+  const token = localStorage.getItem(MicroFrontendToken);
+  if (token) {
+    const parsedToken = JSON.parse(parseJwt(token));
+    if (parsedToken.role) {
+      return parsedToken.role ?? 'default';
+    }
+  }
+
+  return 'default';
+};
```
joshdimanteto, line 40: Line 37 doesn't make sense, look at the if statement above

src/systems/systems.component.tsx
```diff
@@ -492,7 +492,7 @@ function Systems() {
       sx: {
         height: table.getState().isFullScreen
           ? '100%'
-          : getPageHeightCalc('64px + 80px + 40px + 47px + 40px'),
+          : getPageHeightCalc('64px + 80px + 40px + 47px + 40px + 8px'),
```
joshdimanteto: what added the 8px ? can you please put it in the comment above

joshdimanteto (changes requested):
The admin icon should look like this in scigateway 

image

The toggle should be inline with the tabs in dev mode 

image

src/common/authRoleStatus.component.tsx
joelvdavies, line 1: There is still a space at the top.
image

cypress/support/commands.ts
```diff
@@ -123,6 +123,20 @@ Cypress.Commands.add('findBrowserMockedRequests', ({ method, url }) => {
   });
 });
 
+/**
+ * Sets the current token in localstorage to one that contains an admin role
+ * The relevant part of the token's payload is: {role: 'admin', userIsAdmin: false}
+ * This is used whereever specific admin functionality needs to be tested
```
joelvdavies:
```suggestion
 * This is used wherever specific admin functionality needs to be tested
```

src/authProvider.component.tsx
```diff
@@ -0,0 +1,48 @@
...
+      const role = getUserRole();
+
+      setAuthorisation({
+        role: role,
+        isPrivilegedUser: privilegedRoles.includes(role),
+      });
+    };
+    setAuthorisationState();
+
+    // if dev mode add event listener for if token in localstorage changes
```
joelvdavies, line 34: Looking at https://stackoverflow.com/questions/63551506/multiple-identical-eventlisteners-on-the-same-element-why, I think this will keep adding listeners if you change the config? The function above might need to be a useCallback instead?

src/authProvider.component.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+      <AuthProvider>
+        <ConfigTest />
+      </AuthProvider>
+    );
+
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('returns correct authorisation state for admin user', async () => {
```
joelvdavies, line 35: Could these use the new ADMIN_ROLE_TOKEN now? I think I saw another place as well. I am guessing it cant be imported for cypress.

joelvdavies (commented):
Tested again while reviewing the backend. Again worked well for me.

Also could you merge in develop at somepoint as well please, I changed the node version on the CI in https://github.com/ral-facilities/inventory-management-system/pull/1559. 🤞 the flakiness I experienced doesnt reappear here, the main thing I expect that might need updating is some timeouts as the tests take a bit longer but hopefully it wont on this or the subsequent PR.

src/authProvider.component.tsx
MatteoGuarnaccia5: Given it only adds the listener in development, is it much of an issue? When doing testing etc I have not noticed a drop in performance or any unwanted consequences

src/authProvider.component.test.tsx
MatteoGuarnaccia5: Yeah for Cypress it only gets used in the commands.ts as well, in the tests they just call the command  if need be

src/authProvider.component.tsx
joelvdavies: True it probably is fine to leave it as you say impact should be minimal. But equally just because it doesnt give unintended consequences now doesnt mean it wont ever and I still dont think it would be considered best practice, so if there is a simple fix it may as well be addressed.

joelvdavies:
image

I think the height needs to be adjusted based on if there is an admin banner to show or not, given that the bottom now has extra space? (Its the same on other pages)

src/utils.tsx
```diff
@@ -81,11 +95,19 @@ export const getSciGatewayPageHeightCalc = (
...
+): string => {
   // SciGateway heights - view tabs (if in development) - additional
   let newAdditional: string | undefined = undefined;
 
   if (isRunningInDevelopment()) newAdditional = '48px';
+
+  if (authBannerActive) {
+    if (newAdditional === undefined) newAdditional = '-20px';
+    else newAdditional += ' + ' + '-20px';
+  }
```
joelvdavies: Had a look into why this is negative because looking at the line above it for the view tabs, it seems it should be positive. Turns out if you go on dev mode and toggle the banner on and off on the systems page the bottom moves meaning it is not accounting for the space added by it properly - it should be stationary. It seems using a value of positive 16 for this here fixes that but then the spacing still isnt correct. That leaves the gap at the bottom again - but I think that is because the spacing has changed in the top bar containing the breadcrumbs? If so that needs to be adjusted instead which I think is the one in `systems.component`.

MatteoGuarnaccia5: Does changing it to +16px effect the other tables in scigateway?

joelvdavies: No idea, didnt check. But I think the main thing is that I got the value of 16 by gettng the height of the grid that contains the top elements (breadcrumbs and banner) and subtracting the value of it without the banner. So the amount by which the banner changes the page height when present is 56px-40px = 16px, so that is the value that should be used in the utils I think at least for the part that is specific to the banner existing or not.

MatteoGuarnaccia5: 16px doesn't fix the table height, on scigateway there are still gaps under the tables. I think this should be resolved in a different issue and relooked at completely

joelvdavies: Yes it doesnt fix it, I said that above - it stops it being different for when there is and isnt a banner. The fix for the table height is probably adjusting it in `systems.component` instead.

MatteoGuarnaccia5: I meant on all the other pages, e.g. catalogue items table view

joelvdavies: I would need to have a look, but isnt the height difference of having or not having the banner the same on all pages, in which case that value will still be needed? As I say the current value means that it actually moves the position of the bottom of the table when you enable/disable admin which is an indication that its not quite right (this is the case for other tables not just systems).

src/common/authRoleStatus.component.tsx
joelvdavies, line 1: Sorry I think I mentioned this in person but not on the PR, the text needs to be black all the time for the banner otherwise it does this
image
which is not good for contrast (its worse on high contrast mode).

joelvdavies (changes requested):
Sorry one lingering issue, heights are good now though.

joelvdavies (dismissed):

joelvdavies:
@MatteoGuarnaccia5 Sorry just found a wider issue relating to https://github.com/ral-facilities/ldap-jwt-auth/issues/295. If we have to revoke access for anyone the refresh token lasts up to week so at the moment they would remain admin. Once that is fixed, I believe the issue I saw before crops up in the form of 
image
Which Joshua thinks could be due to a similar issue on the frontend of the context not updating when the token is refreshed. Would you be able to create an issue and look into this in a separate PR please?

MatteoGuarnaccia5:
Closing this PR due to git issues, new PR which uses the same branch is #1596

Issue #1523, opened 2025-09-29
add admin functionality for bypassing rules


MatteoGuarnaccia5:
Following on from #1472 , implement functionality to allow user's with privileged roles to bypass rules when

- [ ] moving
- [ ] creating
- [ ] deleting

items

joelvdavies:
Closed by https://github.com/ral-facilities/inventory-management-system/pull/1597

PR #1529, opened 2025-10-06
Implement authorisation for bypassing rules #1523

### LINKED ISSUE #1472: Split Admin Page into "Settings" and "Admin Settings"

MatteoGuarnaccia5:
## Description

see #1472 

Note this PR follows on from #1518 but does not need to wait on any PRs in ims-api or ldap

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

closes #1472


src/items/itemsTable.component.tsx
```diff
@@ -589,12 +598,28 @@ export function ItemsTable(props: ItemTableProps) {
...
+            startIcon={<AddIcon />}
+            sx={{ mx: 0.5 }}
+            variant="outlined"
+            onClick={() => {
+              setItemsDialogType('create');
+              setOpenDialogAsPrivilegedUser(true);
+              table.setCreatingRow(true);
+            }}
+          >
+            Add Item as admin
```
joelvdavies: Looking at the capitalisation perhaps Admin should be capitalised too?
image

src/items/deleteItemDialog.component.tsx
```diff
@@ -72,13 +75,42 @@ const DeleteItemDialog = (props: DeleteItemDialogProps) => {
...
         <WarningIcon sx={{ marginRight: 1 }} />
-        Delete Item
+        Delete Item{isPrivilegedUser ? ' as admin' : ''}
+        {isPrivilegedUser && (
+          <Tooltip
+            title={
+              'As an admin, you can bypass rules that prevent other users from deleting an item'
+            }
+            disableHoverListener={false}
+            aria-label={'admin-status-tooltip'}
```
joelvdavies:
```suggestion
            data-testid={'admin-status-tooltip'}
```
Is this just used for tests? I don't think this is a good aria label for accessibility. Same for any others.

src/items/itemDialog.component.tsx
```diff
@@ -1180,13 +1213,35 @@ function ItemDialog(props: ItemDialogProps) {
...
+            aria-label={'admin-status-tooltip'}
+          >
+            <IconButton
+              disableRipple
+              sx={{ backgroundColor: 'transparent' }}
+              size="large"
+            >
+              <InfoOutlinedIcon
+                sx={(theme) => ({
+                  color: theme.palette?.info.dark,
```
joelvdavies: As mentioned, you may want to use the same colour as the open data label here too so they all match. I also appears a bit off center
image

src/items/itemDialog.component.tsx
```diff
@@ -310,9 +323,23 @@ function ItemDialog(props: ItemDialogProps) {
...
 
   React.useEffect(() => {
-    if (parentSystemId !== selectedItem?.system_id)
+    if (parentSystemId !== selectedItem?.system_id) {
       clearErrorsPropertiesStep('root.formError');
-  }, [clearErrorsPropertiesStep, parentSystemId, selectedItem?.system_id]);
+
+      // Clears usage status error even when they pick a system with no defined rule
+      // which can only be done by privileged users, therefore the usage status dropdown
+      // shows the correct/no error message
```
joelvdavies: Did you mean the correct no error message, I am a bit confused by this comment is it one or the other.

src/items/itemsLandingPage.component.tsx
```diff
@@ -22,23 +22,32 @@ import PrimaryImage from '../common/images/primaryImage.component';
...
+      showAdminEdit={isPrivilegedUser}
       ariaLabelPrefix="items landing page"
       printMenuItem
       uploadAttachmentsEntityId={item.id}
       uploadImagesEntityId={item.id}
       editMenuItem={{
-        onClick: () => setEditItemDialogOpen(true),
+        onClick: (asPrivilegedUser) => {
+          setEditItemDialogOpen(true);
+          setOpenDialogAsPrivilegedUser(asPrivilegedUser === true);
```
joelvdavies, line 49: Is this state needed? Could you not just use `isPrivilegedUser={asPrivilegedUser === true}` in the dialogue?

joelvdavies (commented):
Havent looked at the tests, works well with the backend 🥳.

src/items/itemDialog.component.tsx
joshdimanteto: image
<br>
image

you can just search up InfoOutlinedIcon in VScode to see the other places that it was used 

The info icons should be consistent throughout the whole website bar the open data label as the orange could potential cause contrasted issues

joshdimanteto (changes requested):
- can you reorder the items table action menu to look like this 
<br>
image

image

- Can you add duplicate as Admin as well please. As this should use the admin add item dialog with prefilled values 

- The usage status should be cleared when a user selects a system where there is no valid rule 
<br>


https://github.com/user-attachments/assets/91121e71-fc27-40d8-ab9b-0bef42f029d0


- The usage statuses aggregate cell shouldn't be mandatory, as this is a quick set for all the items within a catalogue item  

image


- The table should default to expanded, as currently you can't see each individual item  

image


- the admin options and normal options should be in the systemsItemsTables, it should be the same as the items table

image

- disable the clearing on the usages status autocompletes, this should allow you to remove all the error validation 
image

src/items/itemsLandingPage.component.tsx
MatteoGuarnaccia5: asPrivileged user isn't accessible in the dialog definition, it's scoped to the onClick, and the user may want to edit as non-admin, I can merge the states together and unify them ?

src/items/itemDialog.component.tsx
MatteoGuarnaccia5: I meant more like the correct (so in this case no) error message. I'll amend the comment as it is confusing

src/items/itemsLandingPage.component.tsx
joelvdavies: Oh right I see now, yeah that was my mistake I confused `asPrivilegedUser` for `isPrivilegedUser`. I guess they could be unified? Would mean there is only one state change rather than two, though I think it should still just be the one render. I place that up to you if you think it would make it any cleaner.

src/systems/systemItemsUsageStatuses.component.tsx
```diff
@@ -0,0 +1,599 @@
...
+  catalogueItem?: CatalogueItem;
+}
+
+export interface SystemItemsUsageStatusTableProps {
+  items: Item[];
+  usageStatuses?: UsageStatusesType[];
+  onChangeUsageStatuses?: (usageStatuses: UsageStatusesType[]) => void;
+  itemUsageStatusesErrorState?: ItemUsageStatusesErrorStateType;
+  onChangeItemUsageStatusesErrorState?: (
+    itemUsageStatusesErrorState: ItemUsageStatusesErrorStateType
```
joshdimanteto: As the values are already pre-polulated we shouldn't need any of the error validation on the table

joshdimanteto (changes requested):

src/systems/systemItemsUsageStatuses.component.tsx
MatteoGuarnaccia5: image

If a user selects a system which does not have a rule, the usage status don't prepopulate. Removing the error logic means the view is like in the image. I still think it is better to have the error messages in the usage status autocompletes, and catalogue item name as previous. As it is obvious to the user which ones are missing, if for example, they only set the first few and forget to scroll down

joshdimanteto: the pre-population function should pre-populate with either the rule value or the previous value if the rule doesn't exist

MatteoGuarnaccia5: I thought you wanted it to be cleared - [see your comment here](https://github.com/ral-facilities/inventory-management-system/pull/1529#:~:text=The%20usage%20status%20should%20be%20cleared%20when%20a%20user%20selects%20a%20system%20where%20there%20is%20no%20valid%20rule)

joshdimanteto: This is a different issue specifically for adding an new item, not edit

joshdimanteto: For edit in the item dialog it shouldn't be cleared, it should keep the current usage status

joshdimanteto (changes requested):
image

- The autocomplete for each item should say usage status instead of usage statuses

image

image

The info icons are not consistence copy to vs edit item dialog as admin. (one is bold and one isn't).

joshdimanteto:
Can you also merge in the latest develop please

MatteoGuarnaccia5:
This PR was incorrectly merged, due to git issues, a new PR which uses the same branch has been made - #1597

PR #1549, opened 2025-10-21
Trim allowed values that end with space #1537

### LINKED ISSUE #1537: Able to add allowed values that end with a space
It is currently possible to create allowed values in a catalogue category property that end (or potentially begin) with a space. When a catalogue item is added or edited this space is removed, causing an error. The values should ideally be trimmed so this case is prevented.

joshdimanteto:
## Description

- Disable trim for string values for the allowed values 

## Testing instructions

- [x] Review code
- [x] Check Actions build
- [x] Review changes to test coverage
- [x] check that allowed values are trim when creating catalogue categories with allowed values (string)
- [x] check that a allowed value selected in the item and catalogue item dialog is not trim when sent back to the backend. when there is white space around it. 

## Agile board tracking

closes #1537


src/catalogue/items/catalogueItemsDialog.component.test.tsx
```diff
@@ -342,6 +345,88 @@ describe('Catalogue Items Dialog', () => {
     });
   }, 10000);
 
+  it('adds a catalogue item where the catalogue item property has an allowed list of values', async () => {
```
joelvdavies: This name is the same as the one above.

joelvdavies (approved):

PR #1559, opened 2025-10-30
Bump node version to v22 on CI to match project & docker #1244

### LINKED ISSUE #1244: Update CI node version to match Dockerfile.prod
Either should modify CI to use node-version: 22, or we should match exactly using a docker file for tests.

joelvdavies:
## Description

Just to close an issue I made a while ago. The latest version is v24, and rennovate is correctly fixing this in that PR (https://github.com/ral-facilities/inventory-management-system/pull/1558) so no idea how it became out of sync. But due to nvm being a bit broken for me, (I cant seem to force the default version and it just resets each time unless I uninstall v22, so may want to fix that/upgrade scigateway too) and because there are some breaking changes looking at the CI, I thought it best to leave it for the moment. Its also only been LTS for 2 days.

I did update to the latest v22 version and fixed the specific version on the CI (I am hoping renovate will keep doing that) so we can even more closely match versions used on the CI and docker images.

This has undone the gains from #1554 unfortunately. CI will fail due to Harbor being down until at least Monday.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

Closes #1244

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
-          node-version: 22
+          node-version: 22.21
           cache: 'yarn'
 
       - name: Install dependencies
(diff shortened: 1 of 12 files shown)


joelvdavies:
4th time in a row 🥳. No idea how considering I didnt change anything to avoid the document is not found errors.

joshdimanteto (approved):

PR #1609, opened 2025-11-13
Add alert banner to items dialog #1607

### LINKED ISSUE #1607: Add alert banner to items dialog

joshdimanteto:
## Description
add alert banner for creation and moving rules when applied 

remove some waits in cypress 

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #1607


cypress/e2e/with_api/items/functions.ts
```diff
@@ -28,9 +28,11 @@ export const modifyItem = (
   cy.findByLabelText('navigate to systems home').click();
 
   if (values.system) cy.findAllByText(values.system).first().click();
-
-  // eslint-disable-next-line cypress/no-unnecessary-waiting
-  cy.wait(4000);
+  if (typeof values.edit === 'number') {
```
joelvdavies: I am confused. What is this edit for?

joelvdavies (commented):
I don't think the dialogue should say moving rule applied when you first go to edit an item and its currently using the same system as its already in, as there is no such thing as a moving rule between the same system, it will just keep the usage status. But dont mind this PR being merged as is if its easier. Worked well for me.

cypress/e2e/with_api/items/functions.ts
joshdimanteto: `edit` gives the index of the table row that is being edited. If it is undefined it assumes that you're adding a new item. I'll changed `edit` to `editIndex`

joshdimanteto:
> I don't think the dialogue should say moving rule applied when you first go to edit an item and its currently using the same system as its already in, as there is no such thing as a moving rule between the same system, it will just keep the usage status. But dont mind this PR being merged as is if its easier. Worked well for me.

I did initially conditionally not display the banner when it has the same system id, but when manual testing it wasn't clear enough and didn't make sense if the user was to navigate a lot through systems and accidentally come back. So it won't be clear why no banner appear for this system, so it would look like a bug to users.

joelvdavies:
> > I don't think the dialogue should say moving rule applied when you first go to edit an item and its currently using the same system as its already in, as there is no such thing as a moving rule between the same system, it will just keep the usage status. But dont mind this PR being merged as is if its easier. Worked well for me.
> 
> I did initially conditionally not display the banner when it has the same system id, but when manual testing it wasn't clear enough and didn't make sense if the user was to navigate a lot through systems and accidentally come back. So it won't be clear why no banner appear for this system, so it would look like a bug to users.

Yeah make sense. I guess in that case the ideal solution to me would be to change the banner text to say this is the same system so there is no rule to apply.

joshdimanteto:
> > I don't think the dialogue should say moving rule applied when you first go to edit an item and its currently using the same system as its already in, as there is no such thing as a moving rule between the same system, it will just keep the usage status. But dont mind this PR being merged as is if its easier. Worked well for me.
> 
> I did initially conditionally not display the banner when it has the same system id, but when manual testing it wasn't clear enough and didn't make sense if the user was to navigate a lot through systems and accidentally come back. So it won't be clear why no banner appear for this system, so it would look like a bug to users.

I am going to alter this though, so the message would be slightly different for root and same system id but that going to be done in a different pr for both `items` and `systemItemsTable` with a warning message

joelvdavies (approved):

PR #1610, opened 2025-11-13
Add alert banner to system items dialog #1608

### LINKED ISSUE #1608: Add alert banner to system items dialog

joshdimanteto:
## Description

add logic to disable the banner if it is system_ids are equal i.e the item hasn't moved 
removes waits in cypress testing

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #1608


src/systems/systemItemsDialog.component.tsx
```diff
@@ -276,6 +278,21 @@ const SystemItemsDialog = React.memo((props: SystemItemsDialogProps) => {
...
+                selectedItems.length !== 0 &&
+                !isSelectedRulesLoading &&
+                !systemsDataLoading && (
+                  <MRTTopTableAlert
+                    title="Item Moving Rule Applied"
+                    showInfoTooltip
+                    infoTooltipTitle={
+                      selectedRules && selectedRules[0]
+                        ? `The items usage statuses will be updated to ${selectedRules[0].dst_usage_status?.value}, as defined by the rules`
+                        : 'The items usage statuses we remain the same, as defined by the rules'
```
joelvdavies:
```suggestion
                        ? `The items usage statuses will be updated to ${selectedRules[0].dst_usage_status?.value}, as defined by the rules`
                        : 'The items usage statuses will remain the same, as defined by the rules'
```
The first of these can probably be the same as the items dialog one or vice versa.  `The item’s usage status will be updated to ${selectedRules?.[0]?.dst_usage_status?.value ?? selectedItem?.usage_status}, as defined by the rules`
The apostrophe used is a wierd one though, might have come from word or something, I get
image

joshdimanteto: This does have to be slightly different as you can move many in systems item dialog and you can only move one in the items dialog

joelvdavies: Thats fine. I think technically it would be `items'` if there are multiple but I wouldnt bother too much as it would depend on if you seleted one or many, what about the wierd character? Also the we needs to be updated as above, it doesnt make sense.

joelvdavies (dismissed):

joelvdavies (dismissed):

joelvdavies (dismissed):

joelvdavies (approved):

PR #1614, opened 2025-11-17
Fix user token not updating #1593

### LINKED ISSUE #1593: User role not updating in token on refresh
A wider issue relating to https://github.com/ral-facilities/ldap-jwt-auth/issues/295. If we have to revoke access for anyone the refresh token lasts up to week so at the moment they would remain admin.

This is believed to be the cause of....

image

As well as the context not updating the users token correctly

_Originally posted by @joelvdavies in https://github.com/ral-facilities/inventory-management-system/issues/1518#issuecomment-3521752231_

MatteoGuarnaccia5:
## Description

see #1593 

A few cases have been investigated for what happens when a user's token contains a privileged role, but the user in the `user_config.yaml` no longer does.

1. User clicks somewhere triggering refresh, their token is now updated with their correct role and `authProvider`'s state recomputes if they are a privileged user
2. User clicks on admin dialog e.g. edit as admin, which triggers refresh of token. The user is able to navigate the dialog but there will be an error stating they are unauthorised / cannot bypass rules so the request is blocked by the backend.
3. User is on admin dialog, then there is a change in the config. The user will be able to submit the dialog as if they were admin as dialog navigation does not trigger a refresh
4. User is on admin dialog, IMS auto refreshes the token (this happens after ~10min), same as case 2 there will be an error following dialog submission

Also changed event name to `token_updated` to follow scigateway formatting and make event name clearer

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

closes #1593


MatteoGuarnaccia5:
Codecov issue is due to lines added in `App.tsx` so I think it can be ignored, a relevant test for the issue has been added

src/App.tsx
```diff
@@ -260,8 +260,10 @@ export function Layout() {
     // attempt to re-render the plugin if we get told to
     const action = (e as CustomEvent).detail;
     if (requestPluginRerender.match(action)) forceUpdate();
-    else if (tokenRefreshed.match(action)) retryFailedAuthRequests();
-    else if (broadcastSignOut.match(action)) clearFailedAuthRequestsQueue();
+    else if (tokenRefreshed.match(action)) {
+      retryFailedAuthRequests();
+      window.dispatchEvent(new CustomEvent('token_updated')); // triggers refresh in authProvider
```
joelvdavies: As these events go across the browser and can be picked up by SciGateway too I think it would make sense to use the same format as SciGateway e.g.
```typescript
export const TokenRefreshedType = 'scigateway:api:token_refreshed';
```
So in this case perhaps `ims:token_updated` or something to make it clear its ims that triggered it. Might be nice to have a this type value defined somewhere too so it can be used like we use `MicroFrontendId`.

src/authProvider.component.test.tsx
```diff
@@ -57,4 +58,25 @@ describe('AuthProvider', () => {
       ).toBeInTheDocument();
     });
   });
+
+  it('re calculates auth state on token update', async () => {
```
joelvdavies:
```suggestion
  it('recalculates auth state on token update', async () => {
```

joelvdavies (commented):
It works! Thanks for debugging it and finding a solution. Just a couple sugggestions.

src/handleIMS_APIError.ts
```diff
@@ -17,6 +17,7 @@ const handleIMS_APIError = (error: AxiosError, broadcast = true): void => {
   // due to the plugin its routing from being different). It is assumed that errors of this nature
   // should not be possible due to SciGateway verifying the user itself, so we follow DataGateway's
   // approach and don't display any of these errors.
+
```
joelvdavies:
```suggestion
```

Not sure why it was addded, comment was meant for the block following the line.

joelvdavies (approved):

Issue #1493, opened 2025-08-05
Use GET endpoints for system type on items landing page and system details


joshdimanteto:
Description:
Currently, the system type is determined by manually finding it from a pre-fetched list of system types.

Expected behavior:
Use a dedicated GET endpoint to retrieve the system type by ID or other identifier on:

- [ ] Items landing page
- [ ] System details page

PR #1651, opened 2026-01-12
Use GET endpoints for system type #1493

lottieo:
## Description

Update items landing page and system details to use GET endpoints for retrieving system type by ID.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

connect to #1493

### DIFF
--- a/src/api/systems.tsx
+++ b/src/api/systems.tsx
@@ -35,19 +35,30 @@ export const getSystemImportanceColour = (
   }
 };
 
-const getSystemTypes = async (): Promise<SystemType[]> => {
-  return imsApi.get(`/v1/system-types`).then((response) => {
-    return response.data;
-  });
+const getSystemTypes = async (
+  system_type_id?: string
+): Promise<SystemType[]> => {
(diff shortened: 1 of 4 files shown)


src/api/systems.tsx
```diff
@@ -35,19 +35,30 @@ export const getSystemImportanceColour = (
...
-  });
+const getSystemTypes = async (
+  system_type_id?: string
+): Promise<SystemType[]> => {
+  const queryParams = new URLSearchParams();
+
+  if (system_type_id) queryParams.append('system_type_id', system_type_id);
+
+  return imsApi
+    .get(`/v1/system-types`, {
```
joelvdavies, line 46:
```suggestion
    .get(`/v1/system-types`, {
```
Sorry, I don't think it was very clear on the issue. It should have mentioned using `/v1/system-type/{ID}` instead of this list endpoint. Like we have a getSystems & useGetSystems as well as singular ones like getSystem & useGetSystem. Right now we get a full list and try to filter out just the one we want, even in cases where we only want one of them.

PR #1652, opened 2026-01-13
Use get endpoints for system type 2 #1493

### LINKED ISSUE #1493: Use GET endpoints for system type on items landing page and system details

lottieo:
## Description
see #1493
## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

Closes #1493


src/api/systems.tsx
joelvdavies, line 1: I only just noticed we have been putting system types in this file and spoke to Joshua about it, could we move them to their own `systemTypes.tsx` file please. We also have a test in `systems.test.tsx` could you move that as well to a similar file. I think for consistency we should probably have a similar unit test for the new function you added here.

joelvdavies (commented):
Works well for me, just one thing I noticed that I think we should do.

joelvdavies (dismissed):
Thanks for moving those, still working well.

Just a reminder to update the PR description, particularly the last line should be `closes` if it closes the issue like it will in this case. Please wait for Joshua to approve as well before merging.

joshdimanteto (changes requested):
Works well. This change also need to be made in the items details panel component as it current get the system type object using the list

lottieo:
Tests currently broken due to issues with harbor

joelvdavies (approved):

joshdimanteto (approved):
LGTM

PR #1677, opened 2026-02-03
add the ims-job-scheduler-api url to config #1676

### LINKED ISSUE #1676: add the ims-job-scheduler-api url to config

joshdimanteto:
## Description

see #1676

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #1676


src/settings.ts
```diff
@@ -3,6 +3,7 @@ import { PluginRoute } from './state/actions/actions.types';
 export interface InventoryManagementSystemSettings {
   imsApiUrl: string;
   osApiUrl: string;
+  jsApiUrl: string;
```
joelvdavies:
```suggestion
  imsJsApiUrl: string;
```
Next to the imsApi one? As unlike the object-storage it has ims in it.

joshdimanteto: I'm more leaning toward getting rid of  `ims`. it seems redundant. also when we add the history `imsHistoryApiUrl` that is too long. 
```
{apiUrl:"",
 osApiUrl:"", 
 jsApiUrl:""
} 
```  
The only one that is important is the docker env var. Everything else you would have to use the wiki docs to know how to setup. so the naming doesn't matter it just need to be consistent within the front end .

joelvdavies: Sure, nice to keep them shorter, I prefer that. I dont fully agree with the last bit, as personally I find its nice to keep things consistent across both the front, backend and deployment where possible as it makes it easy at a glance rather than looking it up for each repo when you are unfamilar.

joshdimanteto: These urls (ones used for the commercialisation) are going to have to work differently anyways. Technically they should be optional and generic. As IMS is not dependent on have the job-scheduler-api. so it should be able to run it the it is not in the config

src/main.tsx
```diff
@@ -127,6 +127,11 @@ export const fetchSettings =
           throw new Error('osApiUrl is undefined in settings');
         }
 
+        // Ensure the jsApiUrl name exists.
+        if (!('jsApiUrl' in settings)) {
+          throw new Error('jsApiUrl is undefined in settings');
+        }
+
```
joshdimanteto, line 134: This is not needed as IMS can run with the jsApi

src/settings.ts
joshdimanteto:
```suggestion
  jsApiUrl?: string;
```

joshdimanteto: I feel like this needs a whole separate conversation anyways on how the actual setup is gonna be, especially on the CI but for now I'll change it to `imsJsApiUrl`

joshdimanteto: #1678

src/main.tsx
joshdimanteto: https://github.com/ral-facilities/inventory-management-system/issues/1678

src/main.tsx
```diff
@@ -210,7 +215,8 @@ async function prepare() {
   if (
     import.meta.env.VITE_INCLUDE_MSW === 'true' ||
     settingsResult?.imsApiUrl === '' ||
-    settingsResult?.osApiUrl === ''
+    settingsResult?.osApiUrl === '' ||
+    settingsResult?.imsJsApiUrl
```
joelvdavies:
```suggestion
    settingsResult?.imsJsApiUrl === ''
```
Odd one out.

joelvdavies (dismissed):

joelvdavies (approved):

Issue #1613, opened 2025-11-17
add notifcation to edit name (in Edit catalogue category dialog)


joshdimanteto:
Currently when you edit the name of a catalogue category does not show  an indicator if it is successful or not. it should display a toast

joelvdavies:
Closed with #1679

PR #1679, opened 2026-02-04
add success toast when updating catalogue category name #1613

MatteoGuarnaccia5:
## Description

see #1613


joshdimanteto (approved):
LGTM

PR #1680, opened 2026-02-04
Catalogue catagories flagging #1659

### LINKED ISSUE #1659: As a user I want to be able to see which catalogue catagories are flagged

joshdimanteto:
## Description
Screenshot 2026-02-04 131117

see #1659

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

close #1659


src/api/api.types.tsx
```diff
@@ -169,6 +169,7 @@ export interface CatalogueCategory
     CreatedModifiedMixin {
   id: string;
   code: string;
+  is_flagged?: boolean;
```
joshdimanteto:
```suggestion
  is_flagged: boolean | null;
```

joshdimanteto:
Add flagging in the catalogue category table aswell

src/utils.tsx
```diff
@@ -691,3 +691,9 @@ export function isExactFilterActive<TData extends MRT_RowData>(
     }
   });
 }
+
+export const criticalityRowStyle = (theme: Theme) => {
+  return {
+    backgroundColor: theme.palette.error.main + '11', // ~7% tint
```
joelvdavies:
```suggestion
    backgroundColor: alpha(theme.palette.error.main, 0.067)
```
I would prefer using MUI's method over appending to a hex string. Feels a bit clearer to me.

src/mocks/CatalogueCategories.json
joelvdavies, line 1: Just curious - Are the mock values going to be accurate i.e. are these flagged when there is something in them that are flagged or are they just going to be randomised?

src/catalogue/category/catalogueCardView.component.tsx
```diff
@@ -150,6 +157,13 @@ const CopyCategoriesButton = (props: {
   );
 };
 
+export const CriticalTooltipText = (
+  <Typography style={{ whiteSpace: 'pre-line' }}>
+    A catalogue category is considered critical if any of its nested child
+    categories or catalogue items are marked as critical.
```
joelvdavies:
```suggestion
    categories or catalogue items are flagged as critical.
```
Would mean it uses the same terminology we have been using.

src/catalogue/category/catalogueCardView.component.tsx
```diff
@@ -304,6 +319,23 @@ function CatalogueCardView() {
         enableColumnFilterModes: false,
         size: 200,
       },
+      {
+        header: 'is Critical',
```
joelvdavies:
```suggestion
        header: 'Flagged as critical',
```
Maybe too long, Otherwise just capitalise the `is`.

cypress/e2e/with_mock_data/catalogueCategories.cy.ts
joelvdavies, line 1: The `adds a catalogue category where isLeaf is true with a list of allowed values` test here doesnt have any mention of flagging in the response yet its still passing. I think the handler should add in the `is_flagged: null` field before returning to match the backend behavior.

src/catalogue/category/catalogueCard.component.tsx
```diff
@@ -43,25 +51,39 @@ function CatalogueCard(props: CatalogueCardProps) {
...
             ? table.options.mrtTheme.selectedRowBackgroundColor
             : undefined,
-        }}
+
+          border: showFlagged
+            ? `2px solid ${theme.palette.error.main}`
+            : undefined,
+
+          boxShadow: showFlagged
+            ? `0 0 10px 3px ${theme.palette.error.main}88` // 88 = ~53% opacity for fuzzy glow
```
joelvdavies:
```suggestion
            ? `0 0 10px 3px ${alpha(theme.palette.error.main, 0.53)}` // 53% opacity for fuzzy glow
```

joelvdavies: I think the shadow makes it look a bit too elevated to me comparing with https://mui.com/material-ui/react-paper/#elevation. Maybe `0 0 5px 2px`? That would look like this:

Image

Instead of
Image

Its personal preference though.

src/catalogue/category/catalogueCategoryDirectoryDialog.component.tsx
joelvdavies, line 1: Not relevant here, just noticed here and preprod has a scrollbar and thought I would mention

Image

On prod the scrollbar is on the dialog, but is about the same size.

joelvdavies (commented):
Minor comments, worked well 🥳. Could you merge develop though to here too please, I think its in the one below this, it kept failing to refresh my token 😄

src/mocks/CatalogueCategories.json
joshdimanteto: random is not worth making it correct in the front end mock data

joelvdavies: Thanks, just wanted to know for reference.

cypress/e2e/with_mock_data/catalogueCategories.cy.ts
joshdimanteto: We don't do this for any of the other extra fields that get added in  catalogue categories such as code, created_time, modified_time etc. It a nice backlog task to get done if there is nothing to do.  I'll create an improve post response issue in ims (#1737)

joshdimanteto: also the cypress tests look at the request body and not the response body so it wouldn't have "is_flagged" anyway

src/utils.tsx
joshdimanteto: why is the suggestion 0.067 instead of just 0.07 ?

joelvdavies: Just that was closer too what the `+ '11'` was doing (17/255=0.666666..). True better to use the 0.7 as its close enough. If you keep the comment then it would also remove the `~` as its exact then.

src/catalogue/category/catalogueCard.component.tsx
```diff
@@ -43,25 +66,43 @@ function CatalogueCard(props: CatalogueCardProps) {
...
+            display: 'flex',
+            flexDirection: 'row',
+            height: '100px',
+            backgroundColor: isSelected
+              ? table.options.mrtTheme.selectedRowBackgroundColor
+              : undefined,
+
+            border: isCriticalMode ? `2px solid ${color}` : undefined,
+
+            boxShadow: isCriticalMode ? `0 0 10px 3px ${color}` : undefined, //  53% opacity for fuzzy glow
```
joshdimanteto:
```
export const criticalityCardStyle = (props: {
  theme: Theme;
  showFlagged: boolean | null;
}) => {
  const { theme, showFlagged } = props;
  const color = getCriticalityColor({ theme, showFlagged });

  return {
    border: `2px solid ${color}`,
    boxShadow: `0 0 10px 3px ${alpha(color, 0.53)}`, // 88 = ~53% opacity for fuzzy glow
  };
};
```
use this

joshdimanteto:
```
      <Card
        sx={(theme) => ({
          width: '100%',
          display: 'flex',
          flexDirection: 'row',
          height: '100px',
          backgroundColor: isSelected
            ? table.options.mrtTheme.selectedRowBackgroundColor
            : undefined,
          ...(isCriticalMode && criticalityCardStyle({ theme, showFlagged })),
        })}
      >
```

src/catalogue/category/catalogueCard.component.tsx
```diff
@@ -14,12 +14,31 @@ import {
...
   table: MRT_TableInstance<CatalogueCategory>;
   card: MRT_Cell<CatalogueCategory>;
 }
 
+export const getCriticalityLabel = (isCritical: boolean | null) => {
+  if (isCritical === true) {
+    return 'This catalogue category is critical.';
+  }
+
+  if (isCritical === false) {
```
joelvdavies, line 35:
```suggestion
  } else if (isCritical === false) {
```

src/api/api.types.tsx
```diff
@@ -170,6 +170,7 @@ export interface CatalogueCategory
     CreatedModifiedMixin {
   id: string;
   code: string;
+  is_flagged: boolean | null;
```
joelvdavies, line 173: Just added a PR, realised we werent doing what I said, we assumed a value of false which was inconsistent with catalogue items, so fixed in https://github.com/ral-facilities/inventory-management-system-api/pull/724.

src/common/__snapshots__/criticalityTooltipIcon.component.test.tsx.snap
joelvdavies, line 1: These snapshots capture the icons, though the text isnt present unless it was hovered over. Lottie went for non snapshot tests with hovers for the spares information tooltip. Could be worth adding the hover, the snapshot probably captures the changing icon better.

src/common/criticalityTooltipIcon.component.tsx
```diff
@@ -0,0 +1,55 @@
...
+  showFlagged: boolean | null;
+}
+
+const getIcon = (showFlagged: boolean | null) => {
+  if (showFlagged === null) {
+    return (
+      <WarningIcon
+        sx={(theme) => ({
+          pr: 1,
+          fontSize: '35px',
```
joelvdavies: Could these all be 36px? Font sizes are usually multiples of 12/even, will mean they are probably 1 pixel off center compared to any text next to them. (Same for below)

src/common/criticalityTooltipIcon.component.tsx
```diff
@@ -0,0 +1,55 @@
...
+          color: theme.palette.warning.main,
+        })}
+      />
+    );
+  }
+
+  if (showFlagged === true) {
```
joelvdavies, line 25:
```suggestion
  } else if (showFlagged === true) {
```

src/mocks/CatalogueCategories.json
```diff
@@ -56,7 +59,8 @@
       }
     ],
     "created_time": "2024-01-01T12:00:00.000+00:00",
-    "modified_time": "2024-01-02T13:10:10.000+00:00"
+    "modified_time": "2024-01-02T13:10:10.000+00:00",
+    "is_flagged": false
```
joelvdavies, line 63: Would be nice to have one or two `null` values since that is a possibility.

src/utils.tsx
```diff
@@ -691,3 +693,28 @@ export function isExactFilterActive<TData extends MRT_RowData>(
...
   });
 }
+
+export const getCriticalityColor = (props: {
+  theme: Theme;
+  showFlagged: boolean | null;
+}) => {
+  const { theme, showFlagged } = props;
+  if (showFlagged === null) return theme.palette.warning.main;
+  if (showFlagged === true) return theme.palette.error.main;
```
joelvdavies, line 703:
```suggestion
  else if (showFlagged === true) return theme.palette.error.main;
```

src/catalogue/category/catalogueCard.component.tsx
joshdimanteto: No need for `else if` here because `return` exits the function immediately. Once a return statement runs, none of the code below it will execute, so using `else` or `else if` would just add unnecessary nesting

src/mocks/CatalogueCategories.json
joshdimanteto: there is one the root of the catalogue

src/common/__snapshots__/criticalityTooltipIcon.component.test.tsx.snap
joshdimanteto: This is already tested in other files ([src/catalogue/category/catalogueCardView.component.test.tsx](https://github.com/ral-facilities/inventory-management-system/pull/1680/changes#diff-b1ca5e35e7b227d06b65cbe7ee5b0e3492ae5090f2c05ba0dcc20cba7b76a75b)[src/catalogue/category/catalogueCardView.component.test.tsx](https://github.com/ral-facilities/inventory-management-system/pull/1680/changes#diff-b1ca5e35e7b227d06b65cbe7ee5b0e3492ae5090f2c05ba0dcc20cba7b76a75b))

src/catalogue/category/catalogueCard.component.tsx
joelvdavies: Sure just felt it looked wierd having multiple ifs like that. It can hint to a compiler that the other code paths exist helping make it faster (switch is actually best for this and is usually preferred over an if when there are many options). But it can be left if you want.

joelvdavies: Seems branch prediction is done through execution rather than specific hints in typescript so likely no difference, ignore me its fine as is.

src/mocks/CatalogueCategories.json
joelvdavies: 🤦‍♂️ Thanks.

src/catalogue/category/catalogueCard.component.tsx
joshdimanteto: Yeah I get what you mean. In this case though, the JS engines don’t really gain anything from using else if. Once you hit a return, the rest of the code is unreachable.

Also, switching to a switch doesn’t really help for a simple true/false/null check. Benchmarks mainly show differences only when you have lots of cases. 

So using plain if statements with early returns is totally fine and keeps the code simple.

joelvdavies (approved):

PR #1717, opened 2026-02-17
Catalogue items flagging #1656

joshdimanteto:
## Description

see #1656 , #1658

## Testing instructions

Add a set up instructions describing how the reviewer should test the code

- [ ] Review code
- [ ] Check Actions build
- [ ] Review changes to test coverage
- [ ] {more steps here}

## Agile board tracking

closes #1656, closes #1658

PR #1738, opened 2026-03-03
Prevent isAdminMode from becoming undefined and ensure false is stored correctly

joshdimanteto:
## Description

This PR fixes an issue where isAdminMode could become undefined after calling setAuthorisation, causing the authorisation listener middleware to attempt storing undefined into local storage.

Key Fixes:

- setAuthorisation now guarantees isAdminMode is always a boolean.
- Prevents accidental writes of undefined to the storage layer.

Restores intended behaviour:

- isAdminMode persists as true or false
- Non‑admin users always reset to false
- Admin users retain their current mode unless explicitly changed

## Testing instructions

- [x] Review code
- [x] Check Actions build
- [x] Review changes to test coverage
- [x]  Admin login retains correct isAdminMode state
- [x]  Non-admin login clears admin mode
- [x]  Middleware no longer writes undefined to storage
- [x]  Hydration logic works correctly for true/false values


joelvdavies (commented):
In the testing instructions you say `Admin login retains correct isAdminMode state` - If I log out as admin then log back in then it gets reset, so that isnt met. I believe because it deletes it on logout. I checked adding the local storage item and then logging in as a non admin and it did clear it so that way round worked as I would hope.

joshdimanteto:
> In the testing instructions you say `Admin login retains correct isAdminMode state` - If I log out as admin then log back in then it gets reset, so that isnt met. I believe because it deletes it on logout. I checked adding the local storage item and then logging in as a non admin and it did clear it so that way round worked as I would hope.

This should be fixed now, Also this pr need to get merged in before #1717 as it fixes a bug i found

joelvdavies (approved):
Working well for me now 🥳

PR #1739, opened 2026-03-04
Allow subsystems to have different type to their parent #1637

### LINKED ISSUE #1637: Allow subsystems to have a different system type to their parent
This issue is the frontend of https://github.com/ral-facilities/inventory-management-system-api/issues/667. Feel free to update with more specific info/split up if needed.

- [x] Allow subsystems to be created with different type to their parent (Could still be autocomplete as most often probably the same as parent? - but want to be able to change)
- [ ] Allow subsystems to have their type changed (when they dont contain items) - no longer just in root systems
- [x] Allow navigation through invalid subsystems when moving items etc (if have a scrapped system located in an operational one, should be able to navigate to it when moving an item from another scrapped system even though it cant be placed within the operational one above it)

joelvdavies:
## Description

See #1637.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

Closes #1637

### DIFF
--- a/src/systems/systemDialog.component.tsx
+++ b/src/systems/systemDialog.component.tsx
@@ -296,15 +296,9 @@ const SystemDialog = React.memo((props: SystemDialogProps) => {
               <Autocomplete
                 disableClearable={value != null}
                 id="systems-type-id-input"
-                disabled={!!parentSystemTypeId}
                 value={
-                  parentSystemTypeId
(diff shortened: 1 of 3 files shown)


joshdimanteto (approved):
LGTM, tested the move to, copy to, moves items and system dialog

PR #1742, opened 2026-03-10
Update eslint-plugin-react-hooks to 7.0.1 and use shareable config

joelvdavies:
## Description

Did while reviewing https://github.com/ral-facilities/datagateway/pull/1878. There is alot more shareble config now so used here like Louise did. Also updates react hooks plugin as it was required for one of these (and is for some reason lumped into the React 19 monorepo update by renovate, when its backwards compatible).

There were two rules I had to disable as I cant see a quick fix for them as they are releated to react hook form & uppy usage, both of which have other updates. Other minor changes were one or two missing dependencies, and some state setters that were used prior to the useState which had to be reordered.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -12,6 +12,16 @@ import globals from 'globals';
 import tseslint from 'typescript-eslint';
 
 export default defineConfig(
+  eslint.configs.recommended,
+  tseslint.configs.recommended,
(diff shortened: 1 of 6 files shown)


joshdimanteto (approved):
LGTM

PR #1744, opened 2026-03-11
manual trigger criticality #1655

### LINKED ISSUE #1655: As an admin I want to be able to manually trigger an update to the catalogue item criticality if its not already running
Create a new card on the settings page that only gets shown to admin. This card should open a dialog which should have that last time it was run and or if it is running. If the criticality is updating the button should be disabled, if not the admin user should be able to re trigger the calculation

joshdimanteto:
## Description

see #1655

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

closes #1655


src/api/api.types.tsx
```diff
@@ -310,3 +310,22 @@ export interface APIImageWithURL extends APIImage {
   view_url: string;
   download_url: string;
 }
+
+// ------------------------------------ JOB SCHEDULER ------------------------------------------------
+
+export enum JobStatus {
+  Running = 'running',
+  Failed = 'failed',
+  Finished = 'finished',
```
joelvdavies, line 319:
```suggestion
  RUNNING= 'running',
  FAILED= 'failed',
  FINISHED= 'finished',
```
I believe the most common convention is to use caps for enum values, though admitedly it seems TypeScript can be a mix (though I know microsoft has a habbit of going against convention like C# vs Java). There are places that dont use it though where the backend is concerned like the system importance we have.

Image

src/api/jobs.test.tsx
joelvdavies, line 1: The name of this file is different, should be `scheduler`. Though to match the naming of the others maybe this should just be jobs.test.tsx, and the other jobs.tsx?

src/mocks/handlers.ts
```diff
@@ -1408,4 +1410,28 @@ export const handlers = [
       );
     }
   ),
+
+  // --------------------------------- JOBS SCHEDULAR ------------------------------------------------------
```
joelvdavies:
```suggestion
  // --------------------------------- JOBS ------------------------------------------------------
```
Would match others?

src/settings/criticality/criticalityJobDialog.component.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+    });
+    await waitFor(() => {
+      expect(screen.queryByRole('progressbar')).not.toBeInTheDocument();
+    });
+    expect(
+      screen.getByText('Job not found. Please contact support.')
+    ).toBeInTheDocument();
+  });
+
+  it('displays warning message, when ims job schedular is not enabled', async () => {
```
joelvdavies:
```suggestion
  it('displays warning message, when ims job scheduler is not enabled', async () => {
```

src/settings/settingsCardView.component.test.tsx
```diff
@@ -15,4 +29,42 @@ describe('SettingsCardView', () => {
...
+  it('renders settings card view correctly no spares definition', async () => {
+    server.use(
+      http.get('/v1/settings/spares-definition', () => {
+        return HttpResponse.json(undefined, { status: 204 });
+      })
+    );
+    const view = createView();
+
+    await waitFor(() => {
+      expect(screen.getByText('Units')).toBeInTheDocument();
```
joelvdavies: Shouldnt this be something other than units?

For me the Criticality card appears and is clickable but doesnt do anything.

joelvdavies (commented):
Worked well all minor comments.

src/api/api.types.tsx
joshdimanteto: I'm going to leave this for now as it used both in PascalCase and UPPER_CASE in IMS, to address in code maintenance

src/settings/settingsCardView.component.test.tsx
joshdimanteto: I have removed this test as the dialog already has the case where the criticality is not set, where it prompts the user to contact support.  This is done to match what the notification does in the next pr

src/api/api.types.tsx
joelvdavies: Could you create an issue for it then please.

src/settings/settingsCardView.component.test.tsx
joelvdavies: Yes I got that one when I turned off the job scheduler, just wasnt sure if it was supposed to be like that when I removed the spares definition?

src/api/jobs.test.tsx
joelvdavies: @joshdimanteto Sorry, just wanted to check I made a couple suggestions on using jobs rather that "job scheduler" as we dont use imsApi, we use systems, catalogue items etc. But not sure if you saw or if you chose not to make the change for some reason as you just resolved the comments rather than replying?

joshdimanteto: Addressed in latest commit

joelvdavies (approved):

src/api/jobs.test.tsx
joelvdavies: Thanks.

PR #1774, opened 2026-03-20
update dev vulnerabilities from github dependabot 

joshdimanteto:
## Description
 ran yarn up tar flatted undici minimatch -R

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3397,12 +3397,12 @@ __metadata:
   languageName: node
   linkType: hard
 
-"brace-expansion@npm:^2.0.1":
-  version: 2.0.1
-  resolution: "brace-expansion@npm:2.0.1"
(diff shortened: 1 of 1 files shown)


joshdimanteto:
I have noticed that we have update undici twice and it has not updated the lockfile  
also in this pr 
https://github.com/ral-facilities/inventory-management-system/pull/1723

joelvdavies:
> I have noticed that we have update undici twice and it has not updated the lockfile also in this pr #1723

Good spot, yes it seems its still pinned by the major version by other subdependencies of `@codecov/vite-plugin`. The plugin itself is not updated much.

joelvdavies (approved):

PR #1790, opened 2026-05-05
fix types on smaller RHF dialogs 

joshdimanteto:
## Description

This form intentionally uses separate types for form state and submit output, aligning with React Hook Form’s TFieldValues and TTransformedValues generics. TFieldValues represents what exists in the form while the user is editing and must match the Zod schema exactly, including requiring file_name so validation correctly errors if the user clears the field.
The Zod schema is therefore the source of truth for the form input shape, ensuring the UI always contains valid, user‑facing values and preventing empty submissions. This keeps validation behavior predictable and avoids silent or surprising outcomes for the user.
TTransformedValues represents what is passed to handleSubmit and ultimately sent to the backend. In this case, that is a PATCH payload where fields are intentionally omitted when unchanged, modeled by ObjectFilePatchBase.
Separating these concerns allows strict UI validation while preserving correct PATCH semantics, without casting or weakening types. This is the intended React Hook Form pattern when form state and persisted data have different shapes.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build


src/systems/systemDialog.component.tsx
```diff
@@ -101,7 +108,7 @@ const SystemDialog = (props: SystemDialogProps) => {
     setError,
     clearErrors,
     reset,
-  } = useForm<SystemPost>({
+  } = useForm<z.input<typeof _systemsPostSchema>, undefined, SystemPost>({
```
joshdimanteto:
```suggestion
  } = useForm<z.input<typeof _SystemsPostSchema>, undefined, SystemPost>({
```

src/manufacturer/manufacturerDialog.component.tsx
```diff
@@ -29,14 +30,20 @@ import { ManufacturerSchema, RequestType } from '../form.schemas';
...
 
-const formControl = createFormControlWithRootErrorClearing<ManufacturerPost>();
 export interface ManufacturerDialogProps {
   open: boolean;
   onClose: () => void;
   selectedManufacturer?: Manufacturer;
   type: RequestType;
 }
 
+const _manufacturerPostSchema = ManufacturerSchema('post');
```
joshdimanteto:
```suggestion
const _ManufacturerPostSchema = ManufacturerSchema('post');
```

PR #1798, opened 2026-05-07
Update node and httpd images in Docker files

joelvdavies:
## Description

See title, addresses some security vulnerabilities.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build


louise-davies (approved):

louise-davies:
Just a thought, might be useful to unpin alpine and instead use a more generic tag (I think I've done this for DGW and OG?), as this means renovate would be able to update these more regularly

joelvdavies:
> ans renovate would be able to update these more regularly

Thanks, yes I did raise it with Viktor, I think we liked being able to see the version directly than having to look up the sha to know what version the image is using. They are infrequent too, so we were fine with the extra burden for the time being.

PR #1808, opened 2026-05-28
Validate spreadsheet for catalogue items #1807

### LINKED ISSUE #1807: validate spreadsheet for catalogue items

joshdimanteto:
## Description

auto validate spreadsheet when it is added to the dialog  
## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

closes #1807


src/common/importTemplateDialog.component.tsx
joelvdavies, line 1: Is this PR validate & ingest - If so I think the issue/pr title should be updated.

joelvdavies, line 1: Can the upload button be disabled when validating? I can get this by clicking the upload quickly

Image

I got an error a while afterwards (though wasnt 100% of the time)

Image

But otherwise no error to say the ingestion failed.

joelvdavies, line 1: If something goes wrong e.g. if the BULK__MAX_CATALOGUE_ITEMS=1 in `ims-api` and left at 1000 on the ingest I get 

Image
There is never a follow up message it just says the validation started, minor though, I would hope it should be obvious from the error itself at the top.

src/common/importTemplateDialog.component.tsx
```diff
@@ -201,6 +213,93 @@ const ImportTemplateDialog = (props: ImportTemplateDialogProps) => {
...
+
+            uppy.info(message, 'error', 15000);
+
+            const filename = `CatalogueItemsValidationErrors-${parentName}.xlsx`;
+
+            handleBlobDownload(response, filename);
+
+            uppy.removeFile(file.id);
+
+            return;
```
joelvdavies, line 248: Is there a reason for all these spaces?

cypress/e2e/with_mock_data/catalogueItems.cy.ts
```diff
@@ -1,7 +1,100 @@
...
+
+  UnsupportedFileExtension: {
+    apiMessage: 'File extension is not supported',
+    uiMessage: 'Content type not supported.',
+  },
+
+  MissingEntity: {
+    apiMessage: 'The specified catalogue category does not exist',
+    uiMessage:
+      'The selected catalogue category no longer exists or is invalid. Please navigate a valid category catalogue that contains catalogue items and try again.',
```
joelvdavies:
```suggestion
      'The selected catalogue category no longer exists or is invalid. Please navigate to a valid category catalogue that contains catalogue items and try again.',
```

cypress/e2e/with_mock_data/catalogueItems.cy.ts
```diff
@@ -1,7 +1,100 @@
...
+export const backendErrorMessage = {
+  IMSIngestAPIVersion: {
+    apiMessage:
+      "Unable to find the custom document property 'IMSIngestAPIVersion' in the workbook properties",
+    uiMessage:
+      'This spreadsheet appears to be corrupted or invalid. Please download a new template and copy your data into it before trying again.',
+  },
+
+  InvalidSpreadsheetFile: {
+    apiMessage: 'File given is not a valid spreadsheet',
```
joelvdavies, line 12:
```suggestion
    apiMessage: 'Spreadsheet file given was not a valid spreadsheet file',
```

cypress/e2e/with_mock_data/catalogueItems.cy.ts
```diff
@@ -1,7 +1,100 @@
...
+    uiMessage:
+      'This spreadsheet appears to be corrupted or invalid. Please download a new template and copy your data into it before trying again.',
+  },
+
+  FileTypeMismatch: {
+    apiMessage:
+      'File does not contain the correct extension or content type do not match',
+    uiMessage: 'File extension does not match content type.',
```
joelvdavies, line 20:
```suggestion
    uiMessage: 'Spreadsheet file extension and content type do not match.',
```

cypress/e2e/with_mock_data/catalogueItems.cy.ts
```diff
@@ -1,7 +1,100 @@
...
+  },
+
+  UnsupportedFileExtension: {
+    apiMessage: 'File extension is not supported',
```
joelvdavies, line 24:
```suggestion
    apiMessage: 'Spreadsheet file extension is not supported',
```

cypress/e2e/with_mock_data/catalogueItems.cy.ts
```diff
@@ -1,7 +1,100 @@
+import { HttpResponse } from 'msw';
+
+export const backendErrorMessage = {
```
joelvdavies, line 3: There is a case of `Cannot have a catalogue items template for a non-leaf catalogue category`, if a catalogue category is changed from leaf to a non leaf after generating a template. Not possible on the frontend, but is allowed via the API provided there are no children.

cypress/e2e/with_mock_data/catalogueItems.cy.ts
```diff
@@ -1,7 +1,100 @@
...
+  MissingSheet: {
+    apiMessage:
+      "Unable to find the 'CatalogueItems Template' sheet in the workbook",
+    uiMessage:
+      'This spreadsheet appears to be corrupted or invalid. Please download a new template and copy your data into it before trying again.',
+  },
+
+  ColumnsModified: {
+    apiMessage:
+      'The columns within the template are either out of date or have been modified.',
```
joelvdavies:
```suggestion
      'The columns within the template are either out of date or have been modified. Please generate a new template.',
```

cypress/e2e/with_mock_data/catalogueItems.cy.ts
```diff
@@ -1,7 +1,100 @@
...
+    uiMessage:
+      'The columns in this spreadsheet are outdated or have been modified. Please download a new template and copy your data into it before trying again.',
+  },
```
joelvdavies, line 86: Also missing the case of "Invalid catalogue item data" - given when the bulk creation on ingest fails due to any of the reasons that would cause validation to fail - most relevant if the category is modified in someway before clicking the ingest button. (Would also occur if the category was made non-leaf between the validate and button click).

src/common/importTemplateDialog.component.test.tsx
```diff
@@ -21,6 +31,57 @@ vi.mock('../utils', async (importOriginal) => {
...
+  TooManyCatalogueItems:
+    'Too many catalogue items in spreadsheet. Found 1050 but only a maximum of 1000 can be processed at once.',
+  CategoryMismatch:
+    'Spreadsheet was generated for a catalogue category with a different ID than the one provided',
+  TemplateVersionMismatch:
+    'Spreadsheet was created by IMS Ingest API v1 which is incompatible with the current version',
+  MissingSheet:
+    "Unable to find the 'CatalogueItems Template' sheet in the workbook",
+  ColumnsModified:
+    'The columns within the template are either out of date or have been modified.',
```
joelvdavies, line 56: Some of these also need updating as above.

src/utils.tsx
```diff
@@ -611,24 +612,107 @@ export const mrtTheme = (theme: Theme): Partial<MRT_Theme> => ({
...
+  if (message.includes('is not supported')) {
+    return 'Content type not supported.';
+  }
+  return 'There was an unexpected error.';
+}
+
+export function parseImageError(message: string): string {
+  if (message.includes('limit for the maximum number of')) {
+    return 'Maximum number of files reached.';
+  }
```
joelvdavies, line 631: Is this supposed to show up somewhere?

I just get 

Image

src/utils.tsx
```diff
@@ -611,24 +612,107 @@ export const mrtTheme = (theme: Theme): Partial<MRT_Theme> => ({
...
+    message.includes(
+      'the columns within the template are either out of date or have been modified.'
+    )
+  ) {
+    return `The columns in this spreadsheet are outdated or have been modified. Please download a new template and copy your data into it before trying again.`;
+  }
+
+  if (message.includes('too many catalogue items in spreadsheet')) {
+    const match = message.match(/maximum of (\d+)/);
+    const max = match ? match[1] : '1000';
```
joelvdavies:
```suggestion
    const max = match ? match[1] : 'Unknown';
```
Just not sure on giving a fallback number when it could be anything.

src/common/importTemplateDialog.component.test.tsx
```diff
@@ -130,4 +197,212 @@ describe('Upload attachment dialog', () => {
 
     expect(await screen.findByText('Complete')).toBeInTheDocument();
   });
+
+  it('displays a warning message when there is warnings in the spreadsheet', async () => {
```
joelvdavies:
```suggestion
  it('displays a warning message when there are warnings in the spreadsheet', async () => {
```

src/common/importTemplateDialog.component.test.tsx
```diff
@@ -130,4 +197,212 @@ describe('Upload attachment dialog', () => {
...
+    expect(xhrPostSpy).toHaveBeenCalledWith(
+      'POST',
+      '/spreadsheets/catalogue-items/ingest',
+      true
+    );
+
+    expect(await screen.findByText('Complete')).toBeInTheDocument();
+  });
+
+  it('displays an error message when there is errors in the spreadsheet', async () => {
```
joelvdavies:
```suggestion
  it('displays an error message when there are errors in the spreadsheet', async () => {
```

src/common/importTemplateDialog.component.test.tsx
```diff
@@ -130,4 +197,212 @@ describe('Upload attachment dialog', () => {
...
+        'Validation failed with 5 errors. A spreadsheet with highlighted issues has been downloaded.'
+      )
+    ).toBeInTheDocument();
+
+    await waitFor(() => {
+      expect(screen.queryByText('Upload 1 file')).not.toBeInTheDocument();
+    });
+  });
+
+  it('displays an error and warning message when there is errors and warnings in the spreadsheet', async () => {
```
joelvdavies:
```suggestion
  it('displays an error and warning message when there are errors and warnings in the spreadsheet', async () => {
```

src/common/importTemplateDialog.component.tsx
joshdimanteto: ingest was in the other pr. as uppy is dealing with the ingest

joshdimanteto: 1. This is not possible using uppy. I don't have access to the button state. 
2. When uppy handles the error, the error message is display if you hover/click over the question mark

joshdimanteto: I'll change this to be a notification within uppy

joshdimanteto: This is consistent to what is done for images and attachments

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joshdimanteto: this would not change as it's the same for images , attachments and images

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joshdimanteto: same as above

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joshdimanteto: That is going to cover by general error case. This shouldn't happened

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joshdimanteto: This is cover by the general error case. I can't explain this error in user friendly way

src/utils.tsx
joshdimanteto: if you click the question mark or hover over it you will see the message

src/utils.tsx
joshdimanteto: image

src/common/importTemplateDialog.component.tsx
joelvdavies: 🤦‍♂️

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joelvdavies: The trouble is that it is different so the backend will return something different even if they are handled in the same way.

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joshdimanteto: https://github.com/ral-facilities/ims-ingest-api/blob/620c2301da73966b531968281628e0615ede7761/ims_ingest_api/routers/spreadsheet.py#L97

joshdimanteto: is your comment correct ? look at link above

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joelvdavies: The general case is this?
image

Its not clear what has happened and as I say it is possible, just not via the frontend right now so it would be nice if it could be added to avoid missing it in future.

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joelvdavies: Same as above - The general error case doesnt show anything right now?

Its really just the that the data is invalid in some way because it was ingested not validated. E.g. "The spreadsheed is invalid. Please revalidate and try again"?

src/utils.tsx
joelvdavies: But that bypasses the bottom one doesnt it? I just thought saying the spreadsheet contains too many items is correct, its just the maximum number wasnt known.

joshdimanteto: if there is no number than they can't fix the problem. we don't want them arbitrarily removes rows if they don't know how many rows to remove

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joshdimanteto: i don't understand what you mean ?

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joshdimanteto: There is no point warning about something that the frontend can't fix. That is support issue

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joshdimanteto: The general error we cover this
image

joshdimanteto: the error message is display if you hover over the question mark

src/common/importTemplateDialog.component.tsx
joshdimanteto: actually i found a way to do this. to block the upload button

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joshdimanteto: This error message covers this case 

https://github.com/ral-facilities/ims-ingest-api/blob/620c2301da73966b531968281628e0615ede7761/ims_ingest_api/routers/spreadsheet.py#L101

which is the same for images, attachments and spreadsheets

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joelvdavies: I dont agree with that statement its not about what the frontend can prevent, its about it dealing with the error cases gracefully in a useful way. Its not a support issue if someone else modifies something after you generated a template.

The general case does say to retry - though retry what part? If they just try uploading the file again they will get the same error then will contact us, which is a waste of time considering it wasnt our problem. Something has changed because of another user and it will take us time to figure that out. I think the message should reflect something to that effect. In this case revalidating isnt enough, they should try generating a new template.

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joelvdavies: Similar to what I said in the above, I dont think this is unexpected though - its not on us to fix, the appropriate course of action is to revalidate the spreadsheet - not contact us for something we cant control.

src/utils.tsx
joelvdavies: But again I dont see how is this an unexpected error - The user should know there spreadsheet contains too many catalogue items - thats the error. The maximum allowed number is useful, and I would hope the regex always works but I dont think the default should be arbitrary.

Perhaps just cut off the second part if it cant figure it out?

src/common/importTemplateDialog.component.tsx
joelvdavies: Thanks, I forgot about the question mark. Fine if it cant be changed, just thought I would raise if it was possible to do something about it.

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joshdimanteto: The issue with this is that a user can't do this using the UI. This is a very specific problem when a programmatic user is using the website at the same time as normal UI user. There is only a 5 min window for this to occur as soon as the data re fetches using tanstack-query the page would automatically update to a catalogue categories page ( hopefully ). The existence of programmatic user doesn't with CLF so this is an unexpected error. In the future i would expect this to change but at that point we would be using i18next (translation file) for our error messages

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joelvdavies: Thats outdated: https://github.com/ral-facilities/ims-ingest-api/blob/add-tests-for-spreadsheets-functionality-%2317/ims_ingest_api/routers/spreadsheet.py#L149. It was purposefully named differently as its using https://github.com/ral-facilities/ims-ingest-api/blob/add-tests-for-spreadsheets-functionality-%2317/ims_ingest_api/schemas/spreadsheets.py#L15-L21. I think I will just change it back, they wont be distinguised by image/attachments when doing https://github.com/ral-facilities/object-storage-api/issues/316, so no reason to make it specific here either.

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joshdimanteto: I'll add the error case, it is just 10 lines

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joelvdavies: Ok reverted now so resolved.

src/utils.tsx
joshdimanteto: This one should be fine like this. it should never happen and if it does something has been missed out or is broken. So we would need to fix it at that point. if the regex consistently doesn't work they would notify us and it would need to be changed

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joelvdavies: It was on the latest branch, though have reverted as mentioned in another comment, so this can be ignored.

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joelvdavies: > The issue with this is that a user can't do this using the UI. This is a very specific problem when a programmatic user is using the website at the same time as normal UI user. There is only a 5 min window for this to occur as soon as the data re fetches using tanstack-query the page would automatically update to a catalogue categories page ( hopefully ). The existence of programmatic user doesn't with CLF so this is an unexpected error. In the future i would expect this to change but at that point we would be using i18next (translation file) for our error messages

Ill agree unlikely, good point about the cache time - I had forgotten, but I still wouldnt say its unexpected to me. Unexpected to the frontend in isolation yes, but not to IMS, and not something to be reported to us.

joshdimanteto: image

@joelvdavies

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joelvdavies: It was in the latest test branch, but have reveted in https://github.com/ral-facilities/ims-ingest-api/pull/20/commits/8ee29018f755e24df8d6dc1326fde00d4247cd4b.

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joshdimanteto: if this error message has multiple case for the same error message e.g. 'The columns within the template are either out of date or have been modified.' and 'Cannot have a catalogue items template for a non-leaf catalogue category' . What should the error message be. I can't instructed the users effectively if there could be multiple different reason why it has failed

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joelvdavies: @joshdimanteto Yeah thats good to me thanks.

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joelvdavies: In all the cases the validation would find errors, for the ingest yes it would raise the same "Invalid catalogue item data" error, but they are all the same message for the frontend not multiple - the only possible cause is when ingesting and ignoring the validation errors or if something has changed. The user's instruction in this case should ideally be to retry the validation as it will tell them what is wrong.

joshdimanteto: image
@joelvdavies

src/utils.tsx
joelvdavies: But we would have no way of knowing its broken? I was just trying to avoid incorrect information from getting presented. Nothing here would break. Is it not just easy to do 

```typescript
const max = match ? match[1] : null;

return `Your spreadsheet contains too many catalogue items.${max ? ` The maximum allowed is ${max}.` : ''} Please reduce the number of rows and try again`;
```
Then it cant go wrong.

cypress/e2e/with_mock_data/catalogueItems.cy.ts
joelvdavies: Looks good to me, thanks.

src/utils.tsx
joshdimanteto: The problem with this is that. if the config was set to 100 and they had 1000 catalogue items in the spreadsheet , the worst case scenario they would have to re upload the spreadsheet 900 times removing 1 catalogue item at a time. The number needs to exist or there is a problem. The regex unlikely to fail but when it does it should be report.   
image

joshdimanteto: this would only fail if the backend error message changes. and at that point that should be a bug fix

joelvdavies: > The problem with this is that. if the config was set to 100 and they had 1000 catalogue items in the spreadsheet , the worst case scenario they would have to re upload the spreadsheet 900 times removing 1 catalogue item at a time. The number needs to exist or there is a problem. The regex unlikely to fail but when it does it should be report. image

Good point, didnt think of that. I like that error message better, thanks.

joelvdavies (commented):
Thanks for the changes, LGTM.

joelvdavies (approved):

Issue #1809, opened 2026-05-28
add comments field to each dialog


MatteoGuarnaccia5:
Part of the history integration work, each dialog now needs to include a comment field to provide the `modified_comment` required by IMS-API

Edit: As part of discussions with clients, the comment field will initially be implemented only for item edit/create and moving

PR #1812, opened 2026-06-02
Add last modified by to table landing pages #1810

MatteoGuarnaccia5:
## Description

Added `modified_by` field onto all table and landing page views. Added it after the `last_modified` columns / text to ensure consistency.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Review changes to test coverage

## Agile board tracking

closes #1810

PR #1816, opened 2026-06-09
Add comment dialog pop items #1809

MatteoGuarnaccia5:
## Description

Adds a modular comment popup dialog, which is integrated with the items dialog and moving items dialog.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

Part of #1809

### DIFF
--- a/cypress/e2e/with_api/items/functions.ts
+++ b/cypress/e2e/with_api/items/functions.ts
@@ -115,6 +115,7 @@ export const modifyItem = (
   }
 
   cy.findByRole('button', { name: 'Finish' }).click();
+  cy.findByRole('button', { name: 'Submit' }).click();
 
   if (!ignoreChecks) {
     if (values.serialNumber) {
@@ -294,6 +295,7 @@ export const duplicateItem = (serialNumber: string, index: number) => {
   cy.findByRole('button', { name: 'Next' }).click();
   cy.findByRole('button', { name: 'Next' }).click();
   cy.findByRole('button', { name: 'Finish' }).click();
+  cy.findByRole('button', { name: 'Submit' }).click();
 
   cy.findAllByText(serialNumber).should('have.length.gte', 2);
 };
--- a/public/res/default.json
+++ b/public/res/default.json
@@ -1,0 +1,30 @@
+{
+  "home_page": {
+    "title_line1": "<0>Inventory Management</0>",
+    "title_line2": "for <1>EPAC</1>",
+    "card1": {
+      "title": "Catalogue",
+      "description": "Browse catalogue categories, sub categories and catalogue items to locate specific items or add new ones.",
+      "link": "/catalogue",
+      "button": "Explore"
+    },
+    "card2": {
+      "title": "Systems",
+      "description": "Navigate through systems and their sub systems to view their component items.",
+      "link": "/systems",
+      "button": "Explore"
+    },
+    "card3": {
+      "title": "Manufacturers",
+      "description": "View the manufacturers of parts used by EPAC.",
+      "link": "/manufacturers",
+      "button": "Explore"
+    },
+    "card4": {
+      "title": "EPAC",
+      "description": "World-leading centre for scientific and industrial research.",
+      "link": "https://www.clf.stfc.ac.uk/Pages/EPAC-Applications.aspx",
+      "button": "Read more"
+    }
+  }
+}
--- a/src/api/api.types.tsx
+++ b/src/api/api.types.tsx
@@ -5,6 +5,10 @@ interface BaseFieldsMixin {
   modified_by: string;
 }
 
+export interface BaseFieldsPost {
+  modified_comment: string | null;
+}
+
 export interface APIError {
   detail: string;
 }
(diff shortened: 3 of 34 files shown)


cypress/e2e/with_api/items/functions.ts
```diff
@@ -294,6 +295,7 @@ export const duplicateItem = (serialNumber: string, index: number) => {
   cy.findByRole('button', { name: 'Next' }).click();
   cy.findByRole('button', { name: 'Next' }).click();
   cy.findByRole('button', { name: 'Finish' }).click();
+  cy.findByRole('button', { name: 'Submit' }).click();
```
joshdimanteto, line 298: I think the overall flow makes sense, but the naming could be a bit confusing from a user perspective. Using “Finish” suggests that the process is complete, but it actually opens another modal where the user still needs to add a comment and submit.

It might be clearer to use something like “Continue” so it better reflects that there is still a final step.

What do you think, does “Finish” accurately reflect the user journey here, or would something like “Continue” make it clearer that there is another step to go through?

public/res/default.json
```diff
@@ -0,0 +1,30 @@
+{
```
joshdimanteto, line 1: This shouldn't be in this pr. I think something has gone wrong when you merged in develop

src/api/api.types.tsx
```diff
@@ -5,6 +5,10 @@ interface BaseFieldsMixin {
   modified_by: string;
 }
 
+export interface BaseFieldsPost {
+  modified_comment: string | null;
+}
+
```
joshdimanteto, line 11: Why is this needed ?

src/api/items.test.tsx
```diff
@@ -61,6 +61,7 @@ describe('items api functions', () => {
             value: false,
           },
         ],
+        modified_comment: null,
```
joshdimanteto, line 64: Can you populate the mock data please it should be a mixture or "null" and different string values

src/history/historyCommentDialog.component.tsx
```diff
@@ -0,0 +1,76 @@
...
+  DialogTitle,
+  Stack,
+  TextField,
+} from '@mui/material';
+
+export interface HistoryCommentProps {
+  open: boolean;
+  onSubmit: (event: React.SyntheticEvent) => void;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  onChange: UseFormRegisterReturn<any>;
```
joshdimanteto, line 17: This could also be used for a dialog that doesn't use react hook form. Refactor this to work for with react hook form and without also you don't need to use any here you can use type generics

src/systems/systemItemsDialog.component.tsx
```diff
@@ -87,6 +89,17 @@ const SystemItemsDialog = React.memo((props: SystemItemsDialogProps) => {
...
 
+  const [historyCommentDialog, setHistoryCommentDialog] =
+    React.useState<boolean>(false);
+
+  // very basic react-hook-form logic to capture modified_comment in popup
+  const { register, watch } = useForm({
+    defaultValues: {
+      modified_comment: null,
+    },
+  });
```
joshdimanteto, line 100: I don't think react-hook-form is needed here. We typically reserve it for more complex forms. Since this dialog only manages a single modified_comment value, useState would be simpler and more readable. Adding react-hook-form for one field introduces extra complexity and mixes form management patterns within the component.

cypress/e2e/with_api/items/functions.ts
MatteoGuarnaccia5: My view on it is, as a user I have entered all the details for my item, I have 'finished' and now I need to 'submit' my new/changed item.

I would leave it as is, if CLF want the wording changed they can tell us what suits their understanding better, otherwise we could change it, then they ask for it to be changed anyways

joshdimanteto: I don't think this is something we should be pushing back to CLF on. From a UX perspective, "Finish" implies the workflow is complete, but clicking it takes the user to another step where they still need to submit.

If there is still another action required after the stepper, then I don't think "Finish" is the correct label. Something like "Continue" or "Review" would more accurately reflect what's actually happening and would be less confusing for users.

MatteoGuarnaccia5: I'm happy to change it to continue where the pop up appears next

src/api/items.test.tsx
MatteoGuarnaccia5: Resolved in #1812

src/systems/systemItemsDialog.component.tsx
MatteoGuarnaccia5: Do we? Dialogs like the unit and usage status that have a single field are also using them

joshdimanteto: Those are different as all of the logic is handled by the react hook form. In this case this partially handle which causes confusion

src/api/api.types.tsx
MatteoGuarnaccia5: Left in by mistake

joshdimanteto (approved):
LGTM

PR #1829, opened 2026-06-19
upgrade alpine images

joshdimanteto:
## Description

update alpine images 
https://hub.docker.com/_/httpd/tags
https://hub.docker.com/_/node/tags

## Testing instructions

- [ ] Review code
- [ ] Check Actions build


joelvdavies (commented):
Might be better to copy this into https://github.com/ral-facilities/inventory-management-system/pull/1775 or otherwise do the node update manually here so the types match.

joelvdavies (approved):

